PR #89, opened 2021-07-01
del trans after the conversion of trans to ndtrans

ykawashima:
I modified moldb.py and exomolapi.py to reduce the memory size by converting trans to ndtrans just after reading the transition data and freeing the memory used for trans.


HajimeKawahara (approved):
Thanks! It's working.

Issue #84, opened 2021-06-24
Some docs questions


bmorris3:
Hi @HajimeKawahara, 

Thanks for your excellent work on this toolkit! I'm beginning to experiment with it, and I have a couple of questions that I couldn't answer with the docs or the paper in my first reads: 

1. *Units*: when running [`AutoRT.spectrum`](http://secondearths.sakura.ne.jp/exojax/exojax/exojax.spec.html#exojax.spec.autospec.AutoRT.spectrum), what are the units of the spectrum? 
2. *Exoplanets*: it's clear that this tool is brilliant for e.g. high resolution brown dwarf spectroscopy in the NIR, but I'm wondering if it's currently applicable to low resolution exoplanet spectroscopy in the optical, where there are many many more spectral lines e.g. from H2O to compute, and most of the resolution information is lost due to low spectral resolution. Are there any shortcuts that we can take to apply this toolkit to retrievals of exoplanet emission spectra, for example?
3. *RAM optimization*: One of the bottlenecks in (2) is loading the line list into memory, where there are >10k lines in a narrow bandpass. Apart from setting `crit` or limiting the range of wavelengths in the calculation, are there any steps we can take to minimize the memory pressure?
4. *Species for Hot Jupiters*: thinking out loud here, could we implement H- opacity as well? I'd be happy to help contribute to this, and would be grateful if you could give me some pointers on how to start.

Thanks again!

HajimeKawahara:
Hello @bmorris3,

Thank you for your interest and relevant questions for exojax!

1.  [`AutoRT.spectrum`](http://secondearths.sakura.ne.jp/exojax/exojax/exojax.spec.html#exojax.spec.autospec.AutoRT.spectrum) uses [`piBarr`](http://secondearths.sakura.ne.jp/exojax/exojax/exojax.spec.html#exojax.spec.planck.piBarr) in it. So the unit should be (erg/s/cm2/cm-1). Note that AutoRT and AutoXS are for a [quick use](http://secondearths.sakura.ne.jp/exojax/userguide/ql.html). The autodiff function is not currently unavailable for them.

[...]

Issue #66, opened 2021-05-26
Hydrogen lines for WD spectra


HajimeKawahara:
as suggested by Kashiyama-san.

HajimeKawahara:
Section 2: https://arxiv.org/pdf/0902.4182.pdf

HajimeKawahara:
Table 1 in 0902.4182.pdf

- H bound-bound, Stark broadening (H–H+, H−e
−) Ni(H) 0902.4182.pdf
- H bound-bound, quasi-molecular (H–H, H–H+) N(H) Allard et al. (2004)
- H bound-bound, neutral broadening (H–H) Ni(H) Ali & Griem (1965, 1966)
- H bound-free Ni(H) Mihalas (1978)
- H free-free NeN(H+) Mihalas (1978)
- H2 free-free NeN(H+2) we assume that of H I
- H3 free-free NeN(H+3) we assume that of H I
- H− bound-free N(H−) John (1988)
- H− free-free NeN(H) John (1988)
- H2+ bound-free N(H+2) Kurucz (1970)
- H2+ free-free N(H)N(H+) Kurucz (1970)
- H2− free-free NeN(H2) Bell (1980)
- CIA H-H2 N(H)N(H2) Gustafsson & Frommhold (2003)
- CIA H2-H2 N(H2)N(H2) Borysow et al. (2001)
- Rayleigh H N(H) Kissel (2000)
- Rayleigh H2 N(H2) Dalgarno & Williams (1962)
- Thompson e scattering Ne Mihalas (1978)

PR #90, opened 2021-07-01
Adding H- continuum opacity from John (1988)

bmorris3:
With some help from @daniel-kitzmann, I've implemented the H- continuum opacity from John (1988). My tests show that this runs super fast (since no interpolation/database is needed).

Do you have any ideas how best to test this, to sanity check that it looks correct? I attach a plot of the absorption coefficient as a function of wavelength, which looks sensible.

hminus

Partly addresses #66, #84.


HajimeKawahara:
Wow! Great! I think the comparison with petitRadtrans would be great. See the last column of[ the available opacity in petitRadtrans](https://petitradtrans.readthedocs.io/en/latest/content/available_opacities.html). For CIA+CO, [this](https://github.com/HajimeKawahara/exojax/blob/master/examples/comparisons/comp_petitRADTRANS_linear.ipynb) is the comparison with petitRadtrans (Note that the grid the profile should be linear in T-logP plain to reduce the effect of interpolation in petitRadtrans) and the code for output by petitRadtran is given [here](https://github.com/HajimeKawahara/exojax/blob/master/examples/comparisons/co_exojax/Gl229_CO_CIA_linear.py).  I think this comparison does not need to include any molecules?

bmorris3:
> I think the comparison with petitRadtrans would be great. 

I agree, this would be great! I'm having some trouble installing pRT though, do you have it ready to run on your machine? If so, would you mind running the sanity check? Sorry to ask for your help but my Mac won't cooperate.

HajimeKawahara:
Got it! I will do that tomorrow morning (I am gonna bring a boy from kindergarten :boy:  ).

HajimeKawahara:
Do you have a specific reason to use number density as input parameters in `dtauHminus`? If not, can I change them to VMR?

HajimeKawahara:
Ah, it's because it is electron.

HajimeKawahara:
Thanks this contribution @bmorris3. I checked it. It looks that hminus absorption is higher than that by petitRadtrans.

The code is [comp_petitRADTRANS_linear_hminus.ipynb](https://github.com/HajimeKawahara/exojax/blob/h-minus/examples/comparisons/comp_petitRADTRANS_linear_hminus.ipynb) in h-minus branch
This is the contribution function of hminus in the current branch assuming MMR_H,e-=0.0001 

cf

And the petitRadtrans ([this](https://github.com/HajimeKawahara/exojax/tree/h-minus/examples/comparisons/petitcode/hminus) is a generation code, output is also in this directory so that you can use it.)

emission_cf

Here I included CO lines, ignore them. 

But, I changed some of your code to compute radiative transfer, it might change the result. It would be great if you could check them.

- moved main function from contdb.py to a new file hminus.py because no database is used for hminus (it's my fault)
- vmap operators are applied to use Tarr, nus as input at [here](https://github.com/HajimeKawahara/exojax/blob/h-minus/src/exojax/spec/hminus.py#L27) L27-L32.
- VMR as input instead of number density because a constant VMR is usually used in planet application. 

Also, another issue is we cannot use for loop for auto-diff of JAX. Instead, [scan](https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.scan.html) can be used or just write down for i=0,7 at [here](https://github.com/HajimeKawahara/exojax/blob/h-minus/src/exojax/spec/hminus.py#L111) and [here](https://github.com/HajimeKawahara/exojax/blob/h-minus/src/exojax/spec/hminus.py#L64). but I just postponed to the edit of them until figuring out the mismatch of the absorption.

Thanks!

daniel-kitzmann:
> Do you have a specific reason to use number density as input parameters in `dtauHminus`? If not, can I change them to VMR?

That is my fault - my original code uses number densities instead of mixing ratios. But it should be pretty straightforward to use mixing ratios instead.

daniel-kitzmann:
> Thanks for this contribution @bmorris3. I've checked it. It looks that the current implementation of the h- absorption is higher than that by petitRadtrans.

We seem to use different literature sources for the opacity. But this line:
https://gitlab.com/mauricemolli/petitRADTRANS/-/blob/master/petitRADTRANS/pyth_input.py#L29
looks suspicious...

HajimeKawahara:
Hello @daniel-kitzmann 
It is true that the uncommented version of petitRADTRANS showed stronger absorption... 

hminus_uncomment

daniel-kitzmann:
> Hello @daniel-kitzmann
> It is true that the uncomment-out version of petitRADTRANS showed stronger absorption...
> 
> hminus_uncomment

I just checked the H- absorption from both literature sources. The different equations they provide yield very comparable absorption cross-sections for both, bound-free and free-free absorption. Any differences we see should be due to the implementation in the different codes. The f2 factor that is commented out in petitRadTrans should in fact don't change the absorption coefficient by a lot...

HajimeKawahara:
Thanks! The difference is not so much. So, there is something wrong in implementation.

HajimeKawahara:
@astrostevanus pointed a typo in petitRadtran

L27 in pyth_input.py

```python
	f2 = 0.  -197.789 + 190.266 * np.log10(lamb_use) - 67.9775*np.log10(lamb_use)**2. \
              + 10.6913*np.log10(lamb_use)**3. - 0.625151*np.log10(lamb_use)*4.
        # -- Once last term was commented out agreement was good. Otherwise opacity was way too large.
```

should be

```python
	f2 = 0.  -197.789 + 190.266 * np.log10(lamb_use) - 67.9775*np.log10(lamb_use)**2. \
              + 10.6913*np.log10(lamb_use)**3. - 0.625151*np.log10(lamb_use)**4.
```

The reference is “The Observation and Analysis of Stellar Photospheres” by David F. Gray.
The difference is the power of 4. After correcting this typo, the uncommented version is not so different from the commented version. 

hminus_modified

So, the implementation in exojax should be close to the above.

daniel-kitzmann:
Interestingly, though, removing the f2 contribution makes the continuum description by Gray more comparable to the John one.

In any case, one difference between the two different implementations might also be the units. The code I gave Brett returns the absorption coefficient in units of cm-1, such that multiplying with a length would yield the optical depth. The code from petitRadTrans seems to work in mass units. Depending on what units the entries in the abundance vector there have, it might be cm2/g or cm-1/g.

bmorris3:
> Also, another issue is we cannot use for loop for auto-diff of JAX. Instead, scan can be used or just write down for i=0,7 at here and here. but I just postponed to the edit of them until figuring out the mismatch of the absorption.

Sorry I misunderstood the way jax works 😅. I'll try fixing this now!

HajimeKawahara:
> The code I gave Brett returns the absorption coefficient in units of cm-1,
Ah, I see. Thanks. This means we should multiply just delta r = H dP/P, where H is the scale height.

```python
    dtauctm = (10**(log_hminus_continuum(wavelength_um, Tarr[:, None], number_density_e, number_density_h)
            +lognarr1[:,None]+lognarr2[:,None]+logkb-logg-logm_ucgs)
            *Tarr[:,None]/mmw*ddParr[:,None])
```
should be changed as 
```python
    dtauctm = (10**(log_hminus_continuum(wavelength_um, Tarr[:, None], number_density_e, number_density_h)
            +logkb-logg-logm_ucgs)
            *Tarr[:,None]/mmw*ddParr[:,None])
```
But It seems there is still deviation between petitRadtran, (now the current implementation gives lower value)

cf

@bmorris3 
Thanks for updating. BTW, I committed several changes to [hminus](https://github.com/HajimeKawahara/exojax/tree/h-minus) branch, but it looks not afffect your branch. Can you use or pull  [this one](https://github.com/HajimeKawahara/exojax/tree/h-minus) ? or do you know how to merge these branches?

bmorris3:
@HajimeKawahara I think I successfully merged in your commits to this PR.

daniel-kitzmann:
I just checked Brett's code with an independent implementation in MATLAB. The bound-free and free-free absorption coefficients are absolutely identical. Your conversion to the optical depth seems to be fine as well, I think.

There is one more important difference between petitRadTrans and Brett's implementation. The latter uses the Saha equation to convert the e- and H abundance to H- for the bound-free absorption. petitRadTrans, on the other hand, takes  directly the H- as an input and determines it abundance somewhere else in the code.

HajimeKawahara:
@daniel-kitzmann  Thanks. That is a very important difference. How can we extract H- abundance in your code?

daniel-kitzmann:
You'll find it here:
https://github.com/bmorris3/exojax/blob/8d9df41d0ffa2cff476041d3705a4ca635e9a9d9/src/exojax/spec/hminus.py#L76

> kappa_bf = jnp.where(
>         (wavelength_um <= lambda_0) & (wavelength_um > 0.125),
>         (0.750 * jnp.power(temperature, -2.5) * jnp.exp(alpha / lambda_0 / temperature) * 
>          (1.0 - jnp.exp( -alpha / wavelength_um / temperature)) * kappa_bf), 
>          0

The kappa_bf at the end is the cross-section in cm2 that has been calculated before.
The term 

> (1.0 - jnp.exp( -alpha / wavelength_um / temperature))

is the stimulated emission correction that is also missing from petitRadTrans (but probably not terribly important)
Finally,

> (0.750 * jnp.power(temperature, -2.5) * jnp.exp(alpha / lambda_0 / temperature) 

is the Saha equation. This one multiplied by the electron pressure in dyn/cm2 and the hydrogen number density in cm-3 yields the H- number density in units of cm-3

Multiplied with the cross-section, this then finally yields the absorption coefficient in cm-1.

HajimeKawahara:
Thank you! Yes, it looks petitRadtrans directly specify H- abundance. But I found that abundances['H-'] in petitRadtrans does not change the contribution function even when assuming zeros for H-. 

For the time being, I would postpone the comparison with petitRadtrans and merge this branch to the develop branch. 
I hope we will confirm that when using H- for real problems, probably when @bmorris3 applies exojax to your work?

daniel-kitzmann:
Here is a plot of the test between the implementation in jax and a separate one in MATLAB (n_h = 1, n_e = 1, T = 3000 K)
Both curves are essentially identical
kappa_ff

However, I noticed one small problem in the jax implementation. For whatever reason, it seems to set the absorption coefficient at the transition point from bound-free to free-free absorption to 0.

daniel-kitzmann:
Hmmm... seems to be a numerical underflow. Are you internally using single-precision variables?

HajimeKawahara:
Thank you for checking. Yes, we need to use single precision variables in JAX basically. This is because of its GPU compatibility.

daniel-kitzmann:
> Thank you for checking. Yes, we need to use single precision variables in JAX basically. This is because of its GPU compatibility.

In that case, you might want to check for zeros before applying the log10. Otherwise, you end up with NaNs for some cases. This usually happens when the densities are small. In such cases the absorption coefficients become smaller than the lower limit of single-precision variables.

HajimeKawahara:
Yeah, definitely I should check for 0 before log10. Thanks.

HajimeKawahara:
Note: big fixed #94

PR #98, opened 2021-07-15
Hdf5

ykawashima:
For the purpose of reducing the required RAM memory when reading numerous transition line data, I have made the following modifications.
- modified moldb.py to save the transition data in hdf file instead of the feather one, which enables us to mask the transition data while reading the transition file after the second time
- To enable the masking when reading the hdf file, I have changed the moldb.py to also save 'updated' nu_lines and calculated Sij0 in the hdf file
- added "mask_needed" parameter in "masking" function so that masking process can be skipped when reading the transition data from the hdf file.

Thank you.


HajimeKawahara:
Thank you!
I confirmed it works for CO (Li2005/Exomol), but after that Exomol looks to started the maintenance of the database....

Issue #16, opened 2021-03-18
No nu_lines in some transition files in ExoMol


HajimeKawahara:
Need support for this case in moldb.

14N-1H3__CoYuTe__06100-06200.trans.bz2

PR #99, opened 2021-07-15
Cut on log of critical value rather than raw value

bmorris3:
Hi @HajimeKawahara and @ykawashima, 

I'm working with the dev branch and had problems querying the new HDF5 tables via the `crit` keyword. I still can't figure out exactly what the cause of the problem is, but the `where` command in `pd.read_hdf` is not respecting the critical `Sij0`. I'm wondering if this has to do with the floats we use to represent small values of Sij0. For example, I've been using TiO with `crit~1e-16` and finding that the `MdbExomol` object that gets returned is empty, even though there should be some lines in my wavelength range. 

In my experimentations, I found that I could fix this if I alter the `where` query based on the _log_ of the Sij0 values, which is what I've implemented in this PR. This makes for a larger HDF5 database, but the tradeoff is that you can then successfully read-in only portions of it. This PR will only "work" if you delete your existing HDF archives.

If you have any insight on how to improve this PR to make the old behavior work correctly, your help is appreciated and I'll happily close this PR 😄

### DIFF
--- a/src/exojax/spec/moldb.py
+++ b/src/exojax/spec/moldb.py
@@ -61,6 +61,7 @@ class MdbExomol(object):
         molecbroad=t0+"__"+self.bkgdatm
 
         self.crit = crit
+        self.log_crit = np.log10(crit)
         self.margin = margin
         self.nurange=[np.min(nurange),np.max(nurange)]
         self.broadf=broadf
@@ -149,6 +150,7 @@ class MdbExomol(object):
 
                 trans["nu_lines"]=self.nu_lines
                 trans["Sij0"]=self.Sij0
+                trans['logSij0']=np.log10(self.Sij0)
                 key="all_nurange"
                 trans.to_hdf(self.trans_file.with_suffix(".hdf"), key=key, format="table", data_columns=True)
                 del trans
@@ -167,7 +169,7 @@ class MdbExomol(object):
                     where.append("nu_lines>nu_lines_min")
                     where.append("nu_lines<nu_lines_max")
                     if not np.isneginf(self.crit):
-                        where.append("Sij0>self.crit")
+                        where.append("logSij0>self.log_crit")
 
                     trans=pd.read_hdf(trans_file.with_suffix(".hdf"), where=where)
                     ndtrans=trans.to_numpy()


bmorris3:
After some more tests, unfortunately this seems not to be working correctly either.

HajimeKawahara:
Hello @bmorris3 !
Can you give us an example that occurs the error when using the current implementation?

bmorris3:
Sure, it looks like this: 

```python
In [1]: import pandas as pd

In [2]: trans=pd.read_hdf('/.database/TiO/48Ti-16O/Toto/48Ti-16O__Toto.trans.hdf', where=['Sij0 > 1e-16'])

In [3]: trans
Out[3]:
Empty DataFrame
Columns: [i_upper, i_lower, A, nu_lines, Sij0, logSij0]
Index: []
```

bmorris3:
OH, this could be the problem! Is it possible that the `crit`/Sij0 values have changed compared to the previous version with the feather implementation? I was using the `crit` that gave me a certain number of lines before, but now it gives me no lines every time. It looks like the maximum Sij0 is different by orders of magnitude compared to the previous implementation.

ykawashima:
Hi @bmorris3 

Thank you so much for pointing this out! This is just from my quick check, but I have realized that when using Toto TiO line list for the first time, the code somwhow yields the following warning when calculating S0, which did not happen when I checked for Li2015 CO line list.

> /home/kawashima/anaconda3/lib/python3.8/site-packages/exojax-0.5.5-py3.8.egg/exojax/spec/exomol.py:24: RuntimeWarning: invalid value encountered in true_divide
  S0=-A*g*np.exp(-hcperk*elower/Tref)*np.expm1(-hcperk*nu_lines/Tref)\
.broad is used.

As this may be related, I will look into it. Sorry.

bmorris3:
No problem, thanks for the quick response!

bmorris3:
It looks like the TiO file has some zero values for the `nu_lines` column: 

```python
In [2]: trans = pd.read_hdf('.database/TiO/48Ti-16O/Toto/48Ti-16O__Toto.trans.hdf')
trans

In [3]: trans
Out[3]:
          i_upper  i_lower             A      nu_lines          Sij0    logSij0
0             601      271  3.482300e-28      0.000000           NaN        NaN
1             602      272  3.592300e-28      0.000000           NaN        NaN
2             603      273  3.712100e-28      0.000000           NaN        NaN
3             604      274  3.831200e-28      0.000000           NaN        NaN
4             605      275  3.961100e-28      0.000000           NaN        NaN
...           ...      ...           ...           ...           ...        ...
58983947   130411   127539  3.845600e-10  30000.011066  1.083500e-44 -43.965171
58983948   165661   162738  2.142300e-05  29999.813031  4.884841e-50 -49.311150
58983949   164872   163548  2.142100e-05  29999.813031  4.884384e-50 -49.311190
58983950   142828   141383  9.992600e-05  29999.997066  1.212398e-65 -64.916355
58983951   142101   142182  9.993300e-05  29999.997093  1.212483e-65 -64.916324
```

This might be causing the problems in the `exomol.Sij0` function which divides by `nu_lines**2` [here](https://github.com/HajimeKawahara/exojax/blob/294e3cd0c104486f66d867b9c910a5cd73944e26/src/exojax/spec/exomol.py#L24-L25).

ykawashima:
Yes, I have also realized that. I found that e_lower and e_upper are calculated to be the same values for some lines for the Toto TiO line list, resulting in zero values for nu_lines. But I have not started to check why this happens. Sorry.

bmorris3:
It looks like this is a problem with the original transition file: 

```python
In [1]: import pandas as pd

In [2]: dat = pd.read_csv('.database/TiO/48Ti-16O/Toto/48Ti-16O__Toto.trans.bz2',compression="bz2",sep="\s+",nam
   ...: es=("i_upper","i_lower","A","nu_lines"))

In [3]: dat
Out[3]:
          i_upper  i_lower             A      nu_lines
0             601      271  3.482300e-28      0.000000
1             602      272  3.592300e-28      0.000000
2             603      273  3.712100e-28      0.000000
3             604      274  3.831200e-28      0.000000
4             605      275  3.961100e-28      0.000000
...           ...      ...           ...           ...
58983947   130411   127539  3.845600e-10  29999.992723
58983948   165661   162738  2.142300e-05  29999.995951
58983949   164872   163548  2.142100e-05  29999.995961
58983950   142828   141383  9.992600e-05  29999.997066
58983951   142101   142182  9.993300e-05  29999.997093

[58983952 rows x 4 columns]
```

Perhaps a simple `where` condition in [exomolapi.read_trans](https://github.com/HajimeKawahara/exojax/blob/294e3cd0c104486f66d867b9c910a5cd73944e26/src/exojax/spec/exomolapi.py#L86) could skip over these lines with zero `nu_lines` values?

bmorris3:
It seems that switching `trans_lines=True` in [`pickup_gE`](https://github.com/HajimeKawahara/exojax/blob/294e3cd0c104486f66d867b9c910a5cd73944e26/src/exojax/spec/exomolapi.py#L131) avoids this problem! Should `trans_lines=True` be the default?

ykawashima:
Glad to hear that it works though I am a bit wondering whether the lines with nu_lines=0 can be just ignored.

As nu_lines information is the optional information as written in Table 12 of [ Exomol's paper](https://arxiv.org/pdf/1603.05890.pdf) (please see also the [explanation](https://github.com/HajimeKawahara/exojax/blob/294e3cd0c104486f66d867b9c910a5cd73944e26/src/exojax/spec/exomolapi.py#L137) at the beginning of pickup_gE @HajimeKawahara wrote), ```trans_lines=False``` is set to be the default as far as I understand.

HajimeKawahara:
Note that the reason why trans_lines=False is default is that some line list in ExoMol does not have information (see #16).

ykawashima:
This is just for your reference, but among all the 58,983,953 transitions in the Toto TiO line list, 3,402 transitions have zero values for the nu_lines in the original transition file. On the other hand, when calculated using the upper and lower energy difference like done [here](https://github.com/HajimeKawahara/exojax/blob/294e3cd0c104486f66d867b9c910a5cd73944e26/src/exojax/spec/exomolapi.py#L170), 5,370 transitions have zero values for nu_lines since the lower and upper state IDs somehow indicate exactly the same energy levels.

I have confirmed that there are 3,249 transitions that satisfy the above both two cases, meaning that 3,402-3,249=153 transitions that have zero values for the nu_lines in the original transition file, but the nu_lines calculated from the upper and lower energy difference are non-zero. I am not sure whether those 153 lines are reliable, but you can include them if you want. This also means that 5,370-3,249=2,121 lines have non-zero values for the nu_lines in the original transition file, but the nu_lines calculated from the upper and lower energy difference are zero.

I think it is better to include a warning or something when nu_lines result in zero values to avoid this kind of this issue. I will work on it. Thank you very much for pointing this out @bmorris3!

HajimeKawahara:
Closed because this issue was fixed by #104

erwanp:
Hello ! Related to your discussion above, I see you're moving to HDF5. 

We're doing a similar move in RADIS https://github.com/radis/radis/pull/306 but reading HDF5 from Vaex instead of Pandas (warning, it's not compatible. VaeX's HDF5 is column-based ; Pandas's PyTables is row-based). Saves memory and also time (may be relevant for @ykawashima). Proof of concept showing how querying/filtering is faster in Vaex can be found here: [test_vaex.ipynb](https://github.com/radis/radis-benchmark/blob/master/manual_benchmarks/test_vaex.ipynb) 

We can discuss it further on the Radis Slack.

ykawashima:
Hello @erwanp! Thank you very much for your useful information. Indeed, I have been worried that reading HDF5 from Pandas and saving Pandas to HDF5 are a bit slow. I will try using Vaex. Thank you very much!!

erwanp:
Keep us updated. @gagan-aryan is also working on similar improvement !

PR #103, opened 2021-07-19
Hdf5

ykawashima:
I have fixed a bug, which happens when reading the transition files of multiple wavenumber ranges. Sorry for this.

### DIFF
--- a/src/exojax/spec/moldb.py
+++ b/src/exojax/spec/moldb.py
@@ -102,20 +102,23 @@ class MdbExomol(object):
                 
         #trans file(s)
         print("Reading transition file")
+        self.Tref=296.0
+        self.QTref=np.array(self.QT_interp(self.Tref))
+
+        where=[]
+        nu_lines_min=self.nurange[0]-self.margin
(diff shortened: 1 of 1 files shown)


src/exojax/spec/moldb.py
```diff
@@ -100,6 +100,17 @@ def __init__(self,path,nurange=[-np.inf,np.inf],margin=1.0,crit=-np.inf, bkgdatm
         self.gQT=jnp.array(pf["QT"].to_numpy()) #grid QT
         self.T_gQT=jnp.array(pf["T"].to_numpy()) #T forgrid QT
                 
+        self.Tref=296.0
+        self.QTref=np.array(self.QT_interp(self.Tref))
+
+        where=[]
+        nu_lines_min=self.nurange[0]-self.margin
+        nu_lines_max=self.nurange[1]+self.margin
```
HajimeKawahara, line 108: should be nurange[-1] ?

HajimeKawahara:
Confirmed. Thanks!

PR #104, opened 2021-07-19
Added a feature to remove transitions with zero wavenumber

ykawashima:
Regarding the issue reported by @bmorris3 in #99, I have modified the code to remove transitions the wavenumbers of which result in zero. This is due to either of the following reasons that are attributed to the original transition/states files.

- The upper and lower state IDs in the transition file indicate the same energy level when referring to the states file
- The value for the wavenumber column in the transition file is zero

I have also added a warning for this case.

### DIFF
--- a/src/exojax/spec/exomolapi.py
+++ b/src/exojax/spec/exomolapi.py
@@ -183,7 +183,7 @@ def pickup_gE(states,ndtrans,trans_file,trans_lines=False):
         jupper=jupper[mask]
         print("WARNING: {0:,} transitions with the wavenumber=zero in {1} have been ignored.".format(len_org - len(nu_lines), trans_file))
         if trans_lines:
-            print("This is because the value for the wavenumber culumn in the transition file is zero for those transitions.")
+            print("This is because the value for the wavenumber column in the transition file is zero for those transitions.")
         else:
             print("This is because the upper and lower state IDs in the transition file indicate the same energy level when referring to the states file for those transitions.")
 
--- a/src/exojax/spec/exomolapi.py
+++ b/src/exojax/spec/exomolapi.py
@@ -128,12 +128,13 @@ def read_states(statesf):
     return dat
 
 
-def pickup_gE(states,ndtrans,trans_lines=False):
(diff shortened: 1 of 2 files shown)


HajimeKawahara:
Thank you! It worked. JFYI, here is the output for [the test code for ToTo](https://gist.github.com/HajimeKawahara/90dcd2e4c51061191bbedaa2f6b96ce6).

```sh
~/exojax/examples/bd(hdf5)>python forward_modit_TiO.py
nugrid is log: mode= modit
Background atmosphere:  H2
Downloading http://www.exomol.com/db/TiO/49Ti-16O/Toto/49Ti-16O__Toto.def
Downloading http://www.exomol.com/db/TiO/49Ti-16O/Toto/49Ti-16O__Toto.pf
Downloading http://www.exomol.com/db/TiO/49Ti-16O/Toto/49Ti-16O__Toto.states.bz2
Downloading http://www.exomol.com/db/TiO/49Ti-16O/49Ti-16O__H2.broad
Error: Couldn't download .broad file and save.
Downloading http://www.exomol.com/db/TiO/49Ti-16O/49Ti-16O__He.broad
Error: Couldn't download .broad file and save.
Downloading http://www.exomol.com/db/TiO/49Ti-16O/49Ti-16O__air.broad
Error: Couldn't download .broad file and save.
Note: Couldn't find the hdf format. We convert data to the hdf format. After the second time, it will become much faster.
Reading transition file
Downloading http://www.exomol.com/db/TiO/49Ti-16O/Toto/49Ti-16O__Toto.trans.bz2
Note: Couldn't find the hdf format. We convert data to the hdf format. After the second time, it will become much faster.
WARNING: 4,017 transitions with the wavenumber=zero in .database/TiO/49Ti-16O/Toto/49Ti-16O__Toto.trans.bz2 have been ignored.
This is because the upper and lower state IDs in the transition file indicate the same energy level when referring to the states file for those transitions.
.broad is used.
Warning: Cannot load .broad. The default broadening parameters are used.
H2-H2
N= 1271
```

PR #112, opened 2021-08-04
Modified to use vaex instead of pandas

ykawashima:
Based on the [kind advice](https://github.com/HajimeKawahara/exojax/pull/99#issuecomment-882553032) from @erwanp, I have modified the code to use vaex, which is indeed much faster!!, instead of pandas. (For example, for the case of 11300-11400 wavenumber grid file of [this Exomol CH4 line list](https://www.exomol.com/data/molecules/CH4/12C-1H4/YT34to10/), the cpu time for the first time call of [moldb.MdbExomol](https://github.com/HajimeKawahara/exojax/blob/22a15577495ab1c2b8f0341307b64d05b7c85075/src/exojax/spec/moldb.py) has been decreased by 70% in my environment. I used nus,wav,res=nugrid(11310,11390,1000) without crit option.)

I switched on the option of convert=True when reading the transition data from the CSV file. This option allows the code to read large CSV file in chunks while taking slightly longer cpu time (for ex., 1.5% for 8300-8400 wavenumber grid file of [the same Exomol CH4 line list](https://www.exomol.com/data/molecules/CH4/12C-1H4/YT34to10/). See [here](https://vaex.io/docs/faq.html) for the details of this option.)

Thank you.

### DIFF
--- a/src/exojax/spec/exomolapi.py
+++ b/src/exojax/spec/exomolapi.py
@@ -125,9 +125,9 @@ def read_trans(transf):
 
     """
     try:
-        dat = vaex.from_csv(transf,compression="bz2",sep="\s+",names=("i_upper","i_lower","A","nu_lines"))
+        dat = vaex.from_csv(transf,compression="bz2",sep="\s+",names=("i_upper","i_lower","A","nu_lines"),convert=True)
     except:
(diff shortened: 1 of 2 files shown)


HajimeKawahara:
Thank you! confirmed.

PR #116, opened 2021-08-08
Benchmark/opacity

HajimeKawahara:
- added benchmark for opacity
- unit test for hjert 
- 20-30 % speed up by modifying special/faddeeva.py


HajimeKawahara:
self merge

PR #117, opened 2021-08-08
Refac/opacity

HajimeKawahara:
- reduced computation time for hjert and ljert in 25%.
- added a unit test for ljert


HajimeKawahara:
self merge

PR #119, opened 2021-08-09
Refac/opacity3

HajimeKawahara:
erfcx refined and test added and mainly refactoring. computational time not so changed though.


HajimeKawahara:
self merge.

PR #123, opened 2021-08-13
Tests/modit

HajimeKawahara:
Refactoring of DIT/MODIT done

Changes:
- I adopted DIT/MODIT with the pre-computation of the contribution and index for wavenumber to a default because I found it was more accurate and as much as fast (slightly faster). The 3D cases are moved to experimental/ditfnu.py and  experimental/moditfnu.py for future dev. 
- now we have three methods for opacity computation (LPF/DIT/MODIT). Each scheme needs some initialization prior to opacity computation. For LPF, we need to make nu matrix ([make_numatrix0](http://secondearths.sakura.ne.jp/exojax/develop/exojax/exojax.spec.html#exojax.spec.make_numatrix.make_numatrix0)), for DIT/MODIT we need to precompute contribution and indexing of wavenumber to the lineshape density matrix. I made a common module for these purposes: [spec.initspec](http://secondearths.sakura.ne.jp/exojax/develop/exojax/exojax.spec.html#module-exojax.spec.initspec). make_numatrix0 is duplicated but decided not to remove. 

Also, I compared LPF and MODIT for CO cross section similar to a comparison of Fig 3. in [the paper](https://arxiv.org/abs/2105.14782).

comparison_modit

Run tests/comparison/modit_CO.py to reproduce this result.


HajimeKawahara:
self merge.

PR #124, opened 2021-08-14
Features/rm feather

HajimeKawahara:
Related to #112. unified to Vaex for large files.
- not only the trans but also the states file is saved as the Vaex format.
- To do so, changed the input of exomolapi.pickup_gE to numpy array of the state from pandas.

### DIFF
--- a/src/exojax/spec/exomolapi.py
+++ b/src/exojax/spec/exomolapi.py
@@ -107,6 +107,7 @@ def read_pf(pff):
 
     """
     dat = pd.read_csv(pff,sep="\s+",names=("T","QT"))
+
     return dat
 
 def read_trans(transf):
@@ -283,6 +284,8 @@ def read_broad(broadf):
                "kalower","kclower","kaupper","kcupper",\
                "v1lower","v2lower","v3lower",\
                "v1upper","v2upper","v3upper"))
+
+
     return bdat
 
 def check_bdat(bdat):
--- a/src/exojax/spec/exomolapi.py
+++ b/src/exojax/spec/exomolapi.py
@@ -147,18 +147,18 @@ def read_states(statesf):
 
     """
     try:        
-        dat = pd.read_csv(statesf,compression="bz2",sep="\s+",usecols=range(4),names=("i","E","g","J"))
+        dat = vaex.from_csv(statesf,compression="bz2",sep="\s+",usecols=range(4),names=("i","E","g","J"),convert=True)
     except:
-        dat = pd.read_csv(statesf,sep="\s+",usecols=range(4),names=("i","E","g","J"))
+        dat = vaex.read_csv(statef,sep="\s+",usecols=range(4),names=("i","E","g","J"),convert=True)
         
     return dat
 
 
-def pickup_gE(states,ndtrans,trans_file,trans_lines=False):
+def pickup_gE(ndstates,ndtrans,trans_file,trans_lines=False):
     """extract g_upper (gup), E_lower (elower), and J_lower and J_upper from states DataFrame and insert them to transition DataFrame.
 
     Args:
-       states: states pandas DataFrame
+       ndstates: states numpy array
        ndtrans: transition numpy array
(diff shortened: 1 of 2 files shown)


erwanp:
Do you have a read/write performance benchmark of Feather vs Hdf5?

HajimeKawahara:
No, but @ykawashima should know.

ykawashima (approved):
Thank you for the update!
I pushed one commit so that only one hdf5 file is produced for the states file. I know this is a bit confusing (sorry), but for the trans file, the code produces the following two hdf5 files;
1. "-.trans.bz2.hdf5", which is produced because of the ```convert=True``` option [here](https://github.com/HajimeKawahara/exojax/blob/30796b870029de996ed3bbd94c0399f0f7609488/src/exojax/spec/exomolapi.py#L128) and is exactly the same as "-.trans.bz2". That option allows the code to read the data even larger than the RAM. See [here](https://vaex.io/docs/faq.html) for details.
2. "-.trans.hdf5" is the trans data with updated nu_lines and Sij0 column added.

Since any modification is not made for the states data, I think it is not needed to export hdf5 file [here](https://github.com/HajimeKawahara/exojax/blob/a7a986a7558596fb25c4905b4b349abf436513c8/src/exojax/spec/moldb.py#L100) in addition to the one produced when reading the states file [here](https://github.com/HajimeKawahara/exojax/blob/665456f9791b5328816c266bb6dcd4659ca41ba8/src/exojax/spec/exomolapi.py#L151) because of the ```convert=True``` option.

ykawashima:
Hi @erwanp. I found that while the cpu time is almost the same or even a little shorter for Feather when the file size is small, for large-size files, time for reading the data from Hdf5 is much shorter compared to Feather. I am sorry that this is just my rough memo when I did the comparison for the Exomol transition files, but here it is. (I did the selection of the transitions with Einstein A coefficient larger than 1e-4 after reading the data from Feather or Hdf5. Also, I am sorry that at that time, I did not know the way to directly convert the data from vaex to numpy while implementing the selection so I used the pandas in between them.)

Exomol CO transition file "12C-16O__Li2015.trans.bz2" (1.4 MB) [here](https://www.exomol.com/data/molecules/CO/12C-16O/Li2015/)
```
Feather
read, bz2=>pd=>numpy 0.23758885264396667 [s]
save, pd=>feather 0.16123626567423344 [s]
read, feather=>pd 0.008673442527651787 [s]

Hdf5
read, bz2=>vaex=>numpy 0.2707656789571047 [s]
save, vaex=>hdf5 0.3378127831965685 [s]
read, hdf5=>vaex=>pd=>numpy 0.5697189848870039 [s]
```

Exomol CH4 transition file "12C-1H4__YT34to10__11900-12000.trans.bz2" (232MB) [here](https://www.exomol.com/data/molecules/CH4/12C-1H4/YT34to10/)
```
Feather
read, bz2=>pd=>numpy 46.019451439380646 [s]
save, pd=>feather 0.8868063762784004 [s]
read, feather=>pd 0.461031774058938 [s]

Hdf5
read, bz2=>vaex=>numpy 46.72854492254555 [s]
save, vaex=>hdf5 0.7241112403571606 [s]
read, hdf5=>vaex=>pd=>numpy 0.5385963618755341 [s]
```

Exomol CH4 transition file "12C-1H4__YT34to10__08300-08400.trans.bz2" (3.6 GB) [here](https://www.exomol.com/data/molecules/CH4/12C-1H4/YT34to10/)
```
Feather
read, bz2=>pd=>numpy 644.1277352403849 [s]
save, pd=>feather 13.745670029893517 [s]
read, feather=>pd 12.460899436846375 [s]

Hdf5
read, bz2=>vaex=>numpy 638.4434926658869 [s]
save, vaex=>hdf5 13.043437561020255 [s]
read, hdf5=>vaex=>pd=>numpy 1.925445280969143 [s]
```

HajimeKawahara:
Thank you for the review!

PR #157, opened 2021-10-05
Update initspec.py

bmorris3:
`initspec.init_dit` expects a `warning` variable to be set, but none is defined. Adding one.


HajimeKawahara:
Thanks!

PR #159, opened 2021-10-08
bugfix: autospec for CN

HajimeKawahara:
Auto recommendation failed. Added the recommendation manually in defmol.py

```python
from exojax.spec import	AutoXS
import numpy as	np
nus=np.linspace(1800.0,2200.0,40000,dtype=np.float64)
autoxs=AutoXS(nus,"ExoMol","CN")
xsv=autoxs.xsection(1000.0,1.0)
```

CN


erwanp:
@minouHub you'll be interested too

PR #176, opened 2021-11-20
Metalline unit test changed

HajimeKawahara:
some change in the unit test for the metal line. 
 
- separated the unit test for different modes into each function in one file.
- check the values not just non nan


HajimeKawahara:
- the data (vald2600.gz) should be "wavelength in vacuum" not in air.  

- Here is just for the record of the messages from @chonma0ctopus on slack (auto translated )

The REFS for vacuum is as follows

REFS=np.array([[1.5435075e-12, 2692.6172, 1742.1697],\
        [1.5435075e-12, 2692.6172, 7155.4468],\
        [1.5435075e-12, 2692.6172, 1281.3076],\
        [1.5435075e-12, 2692.6172, 534.47742],\
        [1.5435075e-12, 2692.6172, 2869.1802]])

If the input file contains absorption lines with wavelengths longer than 100000 Å, the input file will not be valid. If the input file contains absorption lines of wavelengths longer than 100000 Å (attached vald2600_longer.gz), the 13 absorption lines in the range of 1.0 cm^-1 in margin will change the value slightly as follows

REFS=np.array([[1.5435074e-12, 2692.6462, 1742.4810],\
        [1.5435074e-12, 2692.6462, 7156.3286],\
        [1.5435074e-12, 2692.6462, 1281.7037],\
        [1.5435074e-12, 2692.6462, 534.54535],\
        [1.5435074e-12, 2692.6462, 2869.5366]])

If we allow for this difference

assert(diff[0]<1.e-11 and diff[1]<1.e-3 and diff[2]<1.e-3 )

Loosen the condition of

assert(diff[0]<1.e-11 and diff[1]<0.1 and diff[2]<1.0 )

It might also be possible to use something like this.
(However, line 7 of capacity_Fe_test.py clearly states "For this test, the request form should be filled as:", so you may assume that the 13 lines after 100000Å are not there, assuming you follow it. But...) (Supplemental information below)
The reason for stopping at 100000 Å is that VALD has Fe I lines up to about 59524 µm (~6e8 Å) (109250 lines in total). This is because I wanted to omit 1e5 Å and beyond, which have almost no effect on the visible to near-infrared spectrum.
Another reason is that VALD3 absorption line data has a limit of 100000 lines per request (Restrictions on extraction size per request), so it is not possible to download all absorption lines in one request. This is one of the reasons. By the way, I did not get a MemoryError when I tried it again this time, so I thought

nus4LL = 1e8/np.array([1e10, 1500.0])

So I tried again this time and didn't get a MemoryError.

REFS=np.array([[1.5435075e-12, 2704.9580, 1977.1711],\
        [1.5435075e-12, 2704.9580, 7632.7607],\
        [1.5435075e-12, 2704.9580, 1627.6121],\
        [1.5435075e-12, 2704.9580, 599.40393],\
        [1.5435075e-12, 2704.9580, 3064.0417]])

In fact, if we compare them in terms of opacity, we can ignore the difference in the near-infrared wavelength range between the case where we consider up to 1e5 Å and the case where we consider up to 1e10 Å, because it is smaller than the difference between them and petitRADTRANS.
For reference, the red line in the lower panel of the attached figure (wavelength around 12000 Å) is the difference between the case up to 1e5 and the case up to 1e10 normalized by the value for the case up to 1e5 (the blue dashed line is the difference between the case up to 1e10 and petitRADTRANS normalized in the same way).
The blue dashed line shows the normalized difference between petitRADTRANS and exojax up to 1e10.

Translated with www.DeepL.com/Translator (free version)

PR #183, opened 2021-12-01
Atom docs

chonma0ctopus:
Added a user guide to use VALD3 atomic data (will be modified additionally later).


HajimeKawahara:
Thanks!

PR #184, opened 2021-12-02
Metalline

chonma0ctopus:
I revised two example notebooks.
While doing so, I found a couple of issues and fixed them as below.

- Replaced np.where with jnp.where in the functions of pressure broadening (atomll.py)
- Enabled the function "read_ExAll" to remove highly ionized (>3) ions from VALD line list
- Added solar abundance of Ne, Ar, Ge (temporary measures; Atomic input may need to be replaced later)


HajimeKawahara:
Thank you! confirmed that the unit test and two jupyter notebooks worked.

PR #185, opened 2021-12-02
Refac/hitemp

HajimeKawahara:
Reducing DRAM for HITEMP


HajimeKawahara:
To reduce the DRAM for HITEMP, I implemented "extract" option in MdbHit. @ykawashima Can you check this new feature? 
A simple test is tests/mdbhit_extract.py

erwanp:
For HITEMP we should compare this approach with `Radis.io.hitemp.fetch_hitemp`. It is Vaex-based and can load the full CO2 database in a few seconds.

HajimeKawahara:
@erwanp definitely your code is faster than this implementation. We should discuss this in a common framework of I/O!

ykawashima (approved):
I have confirmed the code works, and the implementation has improved the read time significantly. Thank you.

HajimeKawahara:
Thank you!

Issue #190, opened 2021-12-09
Systematics at the edges of the wavenumber grid using MODIT for a lot of lines


HajimeKawahara:
@ykawashima raised this issue on Slack. When computing the cross section using the MODIT and CH4/ExoMol, Yui found a significant increase of the cross section at the edges of the wavenumber grid, such as 

![example](https://user-images.githubusercontent.com/15956904/145415139-f374b9dd-9f77-465f-97a1-fa859340b09e.png)
 
I found the reason of this phenomena. First, when picking the lines up, moldb has margin=1.0 nm as default value outside the wavenumber grid. This implementation was for the direct LPF considering the lines outside of the wavenumber bin leak in the wavenumber grid.

But, for MODIT, it causes some problems: i.e. the lines whose the line center is outside of the wavenumber bin contirbute to the opacity at the edge of the wavenumber bins.  See the line strength distribution of CH4:

![Sij](https://user-images.githubusercontent.com/15956904/145416396-4b96e9d6-9fa6-4f9f-af46-3b7d93ea6bd3.png)

The wavenumber grid was 11890-11950AA. Then lines in 11889-11890AA contributed to the left edge and those in 11950-11951AA  did to the right edge. 

So, I fixed this problem by

- set default margin=0.0 in moldb
- added Warning when some of the line center is outside of the wavenumber grid

But, still be careful for the results. In this implementation, we intentionally ignore the contribution of the lines outside of the wavenumber bins. So, you might use the wavenumber grid larger than your interest. So, in this case, the result becomes like 

![example2](https://user-images.githubusercontent.com/15956904/145416289-58574c5b-3ad9-47d9-bd52-743544dc2b57.png)

PR #191, opened 2021-12-09
removed the margin to prevent the extra contirbutions on the edge of …

HajimeKawahara:
…the wavenumber for MODIT.

This PR solves #190.


ykawashima (approved):
Thank you very much for resolving this issue quickly. I have confirmed that the problem is now solved and the warning works.

HajimeKawahara:
Thank you!

PR #195, opened 2022-01-04
Features/fastchem2

astrostevanus:
Adding fastchem2_call.py in the src/exojax/atm/, fixing a bug in atmprof_Guillot, adding Using FastChem2 to calculate the chemical abundances.ipynb for an example of using FastChem2


HajimeKawahara:
@astrostevanus Thank you! Can you add a google style comments in the source file?

```python
def Hatm(g,T,mu):
    """pressure scale height assuming an isothermal atmosphere
    
    Args:
        g: gravity acceleration (cm/s2)
        T: isothermal temperature (K)
        mu: mean molecular weight
        
    Returns:
        pressure scale height (cm)
    
    """
```
, which will be used to generate [API](http://secondearths.sakura.ne.jp/exojax/exojax/exojax.html).

src/exojax/atm/fastchem2_call.py
```diff
@@ -0,0 +1,169 @@
+#Several functions to simplify the integration of FastChem2 to Exojax, adopted from https://github.com/exoclime/FastChem/blob/master/python/
+# Dictionary of the name of chemical species available in the FastChem2
+# fc_dict=np.array(["P(bar)", "T(k)", "n_<tot>(cm-3)","n_g(cm-3)", "m(u)", "e-", "Al", "Ar", "C", "Ca", "Cl",
```
HajimeKawahara, line 3: you can put the data file in src/exojax/data/.
Do not forget add the data file name in [MANIFEST.in](https://github.com/HajimeKawahara/exojax/blob/master/MANIFEST.in).

HajimeKawahara: [This ](https://github.com/HajimeKawahara/exojax/blob/6bde59cffee96d813275c2464d13422a61f7bf53/src/exojax/spec/atomllapi.py#L291) is an example to read data file in MANIFEST.in.
Another examples is [here.](https://gist.github.com/HajimeKawahara/862428078a9ef412f6ea16fd0131ffaf)

astrostevanus: Thanks! I will add google style comments in the source file and move the data file

HajimeKawahara:
Thanks!

PR #204, opened 2022-01-31
Pyformat

HajimeKawahara:
applied [pyformat](https://pypi.org/project/pyformat/).
All of the tests passed.


HajimeKawahara:
If no one objects, I will merge it later.

Issue #211, opened 2022-02-15
Custom JVP rotational kernel


HajimeKawahara:
I found that response.rigidrot prevented to do jaxopt optimization #72 and also backward autodiff #67. 
So, I made a custom JVP for the rotational kernel. This issue reports on that. 

The derivatives of the rotational kernel are given by  
![image](https://user-images.githubusercontent.com/15956904/154014141-c90cb293-1809-4fc1-b8c2-2f04a2f0a7a1.png)

HajimeKawahara:
Note:

> Even though we defined only a JVP rule and no VJP rule, we can use both forward- and reverse-mode differentiation on f. 

See [here](https://jax.readthedocs.io/en/latest/notebooks/Custom_derivative_rules_for_Python_code.html).

Issue #72, opened 2021-06-04
example for optimization


HajimeKawahara:
So far, we focus on HMC-NUTS as an application of exojax. But, we can use exojax for gradient-based optimization.

HajimeKawahara:
optax

- chex does not work for jaxlib 0.1.60
- it works for jaxlib 0.1.64

HajimeKawahara:
For the optimization of the Voigt profile, see here:
http://secondearths.sakura.ne.jp/exojax/tutorials/optimize_voigt.html

available since v0.5.5

HajimeKawahara:
ExoJAX should be compatible to [jaxopt](https://github.com/google/jaxopt).

HajimeKawahara:
[JAXopt](https://github.com/google/jaxopt) works for ExoJAX emission spectrum anyway. The following is an example of the fitting using [jaxopt.GradientDescent](https://jaxopt.github.io/stable/_autosummary/jaxopt.GradientDescent.html). 
![gradient_descent_jaxopt](https://user-images.githubusercontent.com/15956904/153967613-bcdf0cd4-6b75-40a7-8219-4a0b01231355.png)

HajimeKawahara:
some parameters (e.g. rotational braodening of vsini) are unstable for optimization, yielding nan quickly. We need to find how to avoid such nan.

HajimeKawahara:
#211 solved the vsini issue. The fitting by JAXopt.GradientDescent looks good.
![gradient_descent_jaxopt](https://user-images.githubusercontent.com/15956904/154021102-feb9b1d8-2c0e-4e5c-abde-ca8b4bf2fea3.png)

HajimeKawahara:
from a farther point
https://user-images.githubusercontent.com/15956904/154577407-ffa61525-1462-4285-b93c-7347f16a7790.mp4

Issue #67, opened 2021-05-26
reverse mode


HajimeKawahara:
Currently exojax does not support custom VJP.

HajimeKawahara:
related post: https://github.com/google/jax/issues/3996

bmorris3:
I'm working on combining exojax with another model that might require reverse-mode diff, so I wonder if we can figure this out. Do you have any clues that might help me start to debug? As I'm sure you've experienced, the traceback when `forward_mode_differentiation=False` doesn't give many hints. 

What I'm seeing is that for reverse-mode diff, I'm getting a matrix reshape error that occurs when trying to turn a matrix of shape `[Nlayers x Nlines]` into one of shape `[Nlayers x Nlines x Nnu]`.

bmorris3:
Ah I think I understand, so if you use [this function](https://github.com/HajimeKawahara/exojax/blob/67d1b6c868d69892d4bbf9e620ed05e432cfe61f/src/exojax/spec/rlpf.py#L153) `rxsmatrix`, instead of the standard `xsmatrix`, the reverse mode diff problem goes away?

HajimeKawahara:
Yeah, that function is VJP custom-defined (not JPV). So, the reverse mode for [jax.experimental.optimizer](https://jax.readthedocs.io/en/latest/jax.experimental.optimizers.html), for instance, works. But, I think it does not work for NUTS in numpyro.

I do not have a useful hint, but [a simpler version of the reproduction](https://gist.github.com/HajimeKawahara/e490c44b66d6c719567835260500d86f) of the situation might be useful.

bmorris3:
Does anything need to change for reverse-mode diff with continuum-opacity only calculations? I have a simple model that works in forward mode but doesn't seem to vary the input parameters in reverse-mode (but this may be my fault).

bmorris3:
Update: it was my fault. Nevermind, sorry for the noise!

HajimeKawahara:
[numpyro.plate](https://num.pyro.ai/en/stable/primitives.html#plate) should be used?

dcmvdbekerom:
Hi, I was wondering if this issue is still open.. I tried your minimal example here https://gist.github.com/HajimeKawahara/e490c44b66d6c719567835260500d86f and both forward and reverse worked for me
jaxlib=0.1.75
jax=0.2.27
numpyro=0.8.0

HajimeKawahara:
[...]

PR #212, opened 2022-02-15
enable optimization and reverse mode

HajimeKawahara:
This PR includes the change of rotational broadening #211, which enabled us 
- to do JAXopt based optimization #72 
See http://secondearths.sakura.ne.jp/exojax/tutorials/optimize_spectrum_JAXopt.html
- to do reverse mode #67 
i.e. both
```python
kernel = NUTS(model_c, forward_mode_differentiation=True)    
```
and
```python
kernel = NUTS(model_c, forward_mode_differentiation=False)
```
works for HMC-NUTS.

1. All the pytest passed for my environment.
2. If you want to test it, use tests/reverse/reverse_lpf.py for instance.


HajimeKawahara:
Thank you!

ykawashima:
I think it would be convenient to include it as it is just one package and one would not get an error when trying to use jaxopt. But this is just my opinion and of course, that error would be obvious.

HajimeKawahara:
Thanks. I agree with you and added jaxopt in 'setup.py'.

PR #214, opened 2022-02-20
Refac1.1 1

HajimeKawahara:
all the tests passed.


HajimeKawahara:
just refactoring.
- rlpf is now unnecessary then removed it.

PR #219, opened 2022-02-22
Multi hitemp (bug fix)

ykawashima:
I have just changed a few lines so that the line lists of HITEMP H2O and CO2 should be saved in the directory "01_HITEMP2010" and "02_HITEMP2010" even if the user specify the directory such as "HITEMP2019" or "HITEMP2020".


HajimeKawahara (approved):
Thank you! I confirmed it works.

PR #218, opened 2022-02-21
MODIT+VALD

chonma0ctopus:
Enable to compute the atomic (& ionic) lines from VALD with MODIT.
The database of VALD lines is automatically divided into different species and after the optical depth of each species is calculated, they are summed up.

PR #216, opened 2022-02-21
Multi hitemp

ykawashima:
I have added a feature to read multiple par files for the cases of HITEMP H2O and CO2 (cf. issue #213 ). I wrote a brief instruction [here](https://github.com/HajimeKawahara/exojax/compare/develop...ykawashima:multi_HITEMP?expand=1#diff-3c4e2654380b7828f2d447506b7b08efb6d4b419b81b50897fb3125866dbacb2). Thank you in advance for your review.

Note: I have made the code create a different directory for each wavenumber range. This is because it seems that hapi.db_begin tries to load all the par files in the given directory.

PR #221, opened 2022-02-23
Release v1.1 (retry)

### LINKED ISSUE #94: Confirmation of H- opacity
We haven't compared the hminus opacity merged in the develop branch  to other codes yet.
See #90 for more details.

HajimeKawahara:
We are preparing for release [v1.1](https://github.com/HajimeKawahara/exojax/milestone/4) . Please review this branch. This branch will be merged to `master` branch probably in March 7th unless any critical issues are found. The main changes from version 1.0 are summarized as 

## new capabilities
- VALD3 reverse modeling and MODIT availability #218  See [this tutorial](http://secondearths.sakura.ne.jp/exojax/tutorials/retvald.html).
image

- the reverse differentiation, which enables us to run an HMC in reverse mode and JAXopt #212 See [this tutorial](http://secondearths.sakura.ne.jp/exojax/tutorials/optimize_spectrum_JAXopt.html).

https://user-images.githubusercontent.com/15956904/154577641-543397e1-6db5-4c16-b16f-4b7c7d688800.mp4

- moldb: HITEMP2010 multiple file load #216 #220 
- moldb: Line strength criterion is computed based on a typical temperature of the system. #223

## bug fix
- H minus opacity bug fixed #94

Any comments welcome. In addition, check [user guide](http://secondearths.sakura.ne.jp/exojax/userguide.html) for new capabilities. I can add your edit by manual if you write anything in this conversation thread,  and make sphinx docs. 

Thanks in advance!


astrostevanus:
I can confirm that the HMC-NUTS for VALD3 works well in my PC.
Screen Shot 2022-02-25 at 12 36 59
Screen Shot 2022-02-25 at 12 37 17
.

src/exojax/spec/exomolapi.py
```diff
@@ -19,85 +18,85 @@ def read_def(deff):
...
+        if deff.stem == '14N-1H3__CoYuTe':
+            maxnu = 20000.0
+        if deff.stem == '1H2-16O__BT2':
+            ntransf = 16
+            maxnu = 30000.0
+            exception = True
+            numinf = np.array([0.0, 250.0, 500., 750.0, 1000., 1500.0, 2000,
+                              2250., 2750., 3500., 4500., 5500., 7000., 9000., 14000., 20000.])
+            numtag = make_numtag(numinf, maxnu)
+
```
astrostevanus, line 61: if deff.stem == '16O-1H__MoLLIST':
            alpha_ref = 0.07
            n_Texp = 0.5

astrostevanus, line 61: if deff.stem == '16O-1H__MoLLIST':
            alpha_ref = 0.07
            n_Texp = 0.5

ykawashima (approved):
I have confirmed that the code works for the calculation of any combination of [cross section/spectrum], [HITRAN/HITEMP/ExoMol], and [LPF/MODIT]. I have a few minor suggestions, which I have noticed during the check. (I will be happy to implement the features if needed.)

1. I sometimes carelessly misdefine the observational wavenumber grid and get NaN values from [``ipgauss_sampling``](https://github.com/HajimeKawahara/exojax/blob/f6a599d3ff664dcd50772116e410bf524ccf3ea8/src/exojax/spec/response.py#L92) for the response-applied spectrum. I wondered if it is convenient to give a warning or error message when some sampling wavenumber points are outside the calculated wavenumber range.
2. It seems there is a case that some lines have zero line strengths, such as for the VALD line list of [HiroyukiIshikawa.4214450.gz](http://secondearths.sakura.ne.jp/exojax/data/). So I would like to suggest changing a default value of crit from -np.inf to be 0 so that the lines with zero strength will not be calculated. Or is there any reason to set that as -np.inf?
3. As for the reading of an ExoMol transition file, I switched on the convert option [here](https://github.com/HajimeKawahara/exojax/blob/f6a599d3ff664dcd50772116e410bf524ccf3ea8/src/exojax/spec/exomolapi.py#L135) so that the transition file with its size larger than the memory one is readable (see [here](https://vaex.io/docs/faq.html)). But this option also creates the hdf5 file with identical data to the original transition file (first hdf5 file). So together with the hdf5 exported [here](https://github.com/HajimeKawahara/exojax/blob/f6a599d3ff664dcd50772116e410bf524ccf3ea8/src/exojax/spec/moldb.py#L166), which is the one that will be used subsequently, this results in two hdf5 files for each transition file. In the current version of the code, the first hdf5 file remains as I did not feel comfortable deleting the file while the user might not notice. But on the other hand, this wastes the user's storage space since the size of the transition files is quite large. If you prefer to delete the first hdf5 file, I will implement that feature. What do you think?

HajimeKawahara:
Thank you for reviewing! @ykawashima That are good points. Here is my opinion.
- warning looks very useful.
- actually no reason for - np.nan. it is no problem to use zero instead.
- For instance, how about making an option for saving files and deleting unnecessary files as default?
Can you implement these modifications?

ykawashima:
@HajimeKawahara Thank you for your opinion. Ok. I will work on the modifications.

HajimeKawahara:
@ykawashima 
Can you add any document on the new crit computation in documents/userguide/sijmask.rst?
I think you have some comparisons. 
(to avoid ExoJAX to be a black-box ⚫ )

ykawashima:
@HajimeKawahara Yes, I will do that. I think we also need to update [API](http://secondearths.sakura.ne.jp/exojax/exojax/exojax.spec.html?#exojax.spec.moldb.MdbExomol) (for the parameters crit, Ttyp, and remove_original_hdf) and I was thinking to ask you.

HajimeKawahara:
Thanks! [API](http://secondearths.sakura.ne.jp/exojax/exojax/exojax.spec.html?#exojax.spec.moldb.MdbExomol) updated.

ykawashima:
Thank you!

erwanp (approved):

astrostevanus (approved):
I approve this changes and confirm that everything works well

PR #222, opened 2022-02-26
Minor revisions to userguide and docstrings

chonma0ctopus:
I've made some revisions, mainly to the parts pointed out in #218.
(By the way, thank you for adding retvald.rst, Kawahara-san.)


HajimeKawahara:
Thanks!

PR #223, opened 2022-03-01
Feature/sij0mask

ykawashima:
I have implemented the feature to do the masking based on the line strengths at a given typical temperature ``Ttyp`` instead of 296 K. You can provide the temperature used for the masking as follows;
``
mdbCO=moldb.MdbExomol(emf,nus,crit=1.e-40,Ttyp=1200.)
``
I have set the default typical temperature Ttyp to be 1000 K. Also, I have modified the code so that it would give an error message when the code detects the hdf5 file created by the old version of exojax since it does not contain the elower column. Thank you in advance.

### DIFF
--- a/src/exojax/spec/moldb.py
+++ b/src/exojax/spec/moldb.py
@@ -45,7 +45,7 @@ class MdbExomol(object):
            nurange: wavenumber range list (cm-1) [min,max] or wavenumber grid
            margin: margin for nurange (cm-1)
            crit: line strength lower limit for extraction
-           Ttyp: typical temperature used to calculate the line strength for the above extraction
+           Ttyp: typical temperature to calculate Sij(T) used in crit
            bkgdatm: background atmosphere for broadening. e.g. H2, He,
            broadf: if False, the default broadening parameters in .def file is used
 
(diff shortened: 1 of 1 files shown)


HajimeKawahara:
@ykawashima Thank you! This modification is slight (and important). So I would include this one in version 1.1. Can you replace the branch to be merged from `develop` to `releasev1.1`?

src/exojax/spec/moldb.py
```diff
@@ -36,14 +37,15 @@ class MdbExomol(object):
...
-    def __init__(self, path, nurange=[-np.inf, np.inf], margin=0.0, crit=-np.inf, bkgdatm='H2', broadf=True):
+    def __init__(self, path, nurange=[-np.inf, np.inf], margin=0.0, crit=-np.inf, Ttyp=1000., bkgdatm='H2', broadf=True):
         """Molecular database for Exomol form.
 
         Args:
            path: path for Exomol data directory/tag. For instance, "/home/CO/12C-16O/Li2015"
            nurange: wavenumber range list (cm-1) [min,max] or wavenumber grid
            margin: margin for nurange (cm-1)
            crit: line strength lower limit for extraction
+           Ttyp: typical temperature used to calculate the line strength for the above extraction
```
HajimeKawahara, line 48: Ttyp: typical temperature to calculate Sij(T) used in `crit`?

src/exojax/spec/moldb.py
```diff
@@ -169,7 +184,13 @@ def __init__(self, path, nurange=[-np.inf, np.inf], margin=0.0, crit=-np.inf, bk
                     cdt = (trans.nu_lines > self.nurange[0]-self.margin) \
                         * (trans.nu_lines < self.nurange[1]+self.margin)
                     if not np.isneginf(self.crit):
-                        cdt = cdt * (trans.Sij0 > self.crit)
+                        if not '_elower' in trans:
+                            print('The hdf5 file for the transition file was created using the old version of exojax. Remove', trans_file.with_suffix('.hdf5'), 'and try again.')
```
HajimeKawahara, line 188: The hdf5 file for the transition file was created using the old version of exojax<1.1.

ykawashima:
Thank you. I have modified the explanation and error comments. Also, I have changed the branch to be merged.

ykawashima:
I have fixed the bug raised by @astrostevanus. Thank you for pointing this out.

HajimeKawahara:
Confirmed it worked. Thanks!

PR #226, opened 2022-03-05
enabled masking based on Ttyp for HITRAN/HITEMP

ykawashima:
As I realized that masking based on the typical temperature was unavailable for HITRAN/HITEMP, I have implemented it. Sorry for the multiple pull requests and thank you.

### DIFF
--- a/src/exojax/spec/moldb.py
+++ b/src/exojax/spec/moldb.py
@@ -440,7 +440,7 @@ class MdbHit(object):
         n_air (jnp array): air temperature exponent
     """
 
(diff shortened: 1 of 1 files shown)


HajimeKawahara (approved):
Thanks! 
Checked 
```sh
    61	21:40	cd comparison/
    62	21:40	python dit_CO.py
    63	21:42	python modit_CO.py
    65	21:42	cd ../xs/
    66	21:42	python mdbhit_extract.py
```
worked.

PR #237, opened 2022-03-13
Refac/moldb

HajimeKawahara:
I just did refactoring of moldb.py and confirmed that all the unit tests passed. Just glance at the lines updated just in case.
In addition, I made a demo test to reproduce [Ttyp tutorial](http://secondearths.sakura.ne.jp/exojax/userguide/moldb.html#masking-weak-lines) in tests/xs/Ttyp_demo.py, for future test (this is not the unit test though).


ykawashima (approved):
Thank you. I have confirmed the modifications.

PR #243, opened 2022-04-06
fixed a bug regarding the tab position

ykawashima:
I have fixed a bug regarding the tab position, which is pointed out by @chonma0ctopus. Thank you in advance.


chonma0ctopus (approved):
Confirmed. Thank you for your prompt revision.

HajimeKawahara (approved):
Thanks!

PR #245, opened 2022-04-14
Setdgm hitran

ykawashima:
I have added a feature to use the setdgm when using HITRAN/HITEMP database. I note that the user needs to provide the reference partial pressure array since HITRAN/HITEMP use that information. Also, the partition function ratio for T > 3,500 K is assumed to be the value at 3,500 K (in that case, the warning will be issued).
Thank you in advance.

### DIFF
--- a/src/exojax/spec/moldb.py
+++ b/src/exojax/spec/moldb.py
@@ -755,9 +755,9 @@ class MdbHit(object):
         for iso in self.uniqiso:
             Tmin, Tmax = hapi.get_TMIN_TMAX_FOR_BD_TIPS_2017_PYTHON(self.molecid, iso)
             if(max(allT) > Tmax):
-                print('Warning: partition function ratio for isotope %d of molecule %d at temperatures > %.1fK is assumed to be the value at %.1fK since HAPI does not support those temperatures.' % (iso, self.molecid, Tmax, Tmax))
+                print('Warning: partition function ratio for isotope #%d of molecule #%d at temperatures > %.1fK is assumed to be the value at %.1fK since HAPI does not support those temperatures.' % (iso, self.molecid, Tmax, Tmax))
             if(min(allT) < Tmin):
-                print('Warning: partition function ratio for isotope %d of molecule %d at temperatures < %.1fK is assumed to be the value at %.1fK since HAPI does not support those temperatures.' % (iso, self.molecid, Tmin, Tmin))
+                print('Warning: partition function ratio for isotope #%d of molecule #%d at temperatures < %.1fK is assumed to be the value at %.1fK since HAPI does not support those temperatures.' % (iso, self.molecid, Tmin, Tmin))
             allT_iso = list(np.clip(allT, Tmin, Tmax))
             Qrx.append(hapi.partitionSum(self.molecid, iso, allT_iso))
         Qrx = np.array(Qrx)
--- a/src/exojax/spec/moldb.py
+++ b/src/exojax/spec/moldb.py
@@ -753,7 +753,13 @@ class MdbHit(object):
         allT = list(np.concatenate([[self.Tref], Tarr]))
         Qrx = []
         for iso in self.uniqiso:
-            Qrx.append(hapi.partitionSum(self.molecid, iso, allT))
+            Tmin, Tmax = hapi.get_TMIN_TMAX_FOR_BD_TIPS_2017_PYTHON(self.molecid, iso)
+            if(max(allT) > Tmax):
+                print('Warning: partition function ratio for isotope %d of molecule %d at temperatures > %.1fK is assumed to be the value at %.1fK since HAPI does not support those temperatures.' % (iso, self.molecid, Tmax, Tmax))
+            if(min(allT) < Tmin):
+                print('Warning: partition function ratio for isotope %d of molecule %d at temperatures < %.1fK is assumed to be the value at %.1fK since HAPI does not support those temperatures.' % (iso, self.molecid, Tmin, Tmin))
+            allT_iso = list(np.clip(allT, Tmin, Tmax))
+            Qrx.append(hapi.partitionSum(self.molecid, iso, allT_iso))
         Qrx = np.array(Qrx)
         qr = Qrx[:, 1:].T/Qrx[:, 0]  # Q(T)/Q(Tref)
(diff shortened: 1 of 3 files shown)


erwanp:
Hello @ykawashima . 

- HAPI (using TIPS 2020) has Partition functions way above 3500 K for certain species.  (CO goes up to 6000 K I think ? )
- In any case, I would strongly advise against using the value of 3500 K if values above 3500K are not defined. Partition function is increasing exponentially with the temperature; you would make critical errors. Better raise an OutOfRange error.

See for instance how it's computed in RADIS https://radis.readthedocs.io/en/latest/auto_examples/calculate_partition_functions_from_tips.html#sphx-glr-auto-examples-calculate-partition-functions-from-tips-py

ykawashima:
Hello @erwanp 

Thank you very much for pointing it out. I did not realize that the temperature supported by HAPI is different for each species. I will make adjustments. Thank you very much!!

ykawashima:
I created a function "get_TMIN_TMAX_FOR_BD_TIPS_2017_PYTHON" in hapi.py to get the minimum and maximum temperature values supported by HAPI for each molecule and isotope. I also modified Qr_HAPI in modlb.py so that it assumes the value of qr for temperatures outside the supported values to be the values at those limit temperatures.

You can call setdgm_hitran something like that.
``setdgm_hitran(mdb,fT,Parr,Parr*1.e-3,R,molmass,res,T0_test,alpha_test)``
I note that you need to provide a reference Pself (Parr*1e-3 in this case).

You can check the warning for example, by setting the maximum temperature > 3500 K for HITRAN H2S and 6000 K for NH3.

Thank you.

src/exojax/spec/moldb.py
```diff
@@ -753,7 +753,13 @@ def Qr_HAPI(self, Tarr):
         allT = list(np.concatenate([[self.Tref], Tarr]))
         Qrx = []
         for iso in self.uniqiso:
-            Qrx.append(hapi.partitionSum(self.molecid, iso, allT))
+            Tmin, Tmax = hapi.get_TMIN_TMAX_FOR_BD_TIPS_2017_PYTHON(self.molecid, iso)
+            if(max(allT) > Tmax):
+                print('Warning: partition function ratio for isotope %d of molecule %d at temperatures > %.1fK is assumed to be the value at %.1fK since HAPI does not support those temperatures.' % (iso, self.molecid, Tmax, Tmax))
```
erwanp: sorry to insist but users could be making an extreme error by doing this. 

Partition functions cannot be clipped. As you can see running the code below, they are diverging fast with temperature :

```python

from radis.levels.partfunc import PartFuncHAPI
import numpy as np

T = np.linspace(300, 9000)

Q = PartFuncHAPI("CO", I=1)
Z = []
for Ti in T:
    Z.append(Q.at(T=Ti))
    
import matplotlib.pyplot as plt
plt.plot(T, Z)
plt.xlabel('Temperature')
plt.ylabel('Partition Function')
```
image



 Having only a print() notice is dangerous --> no-one reads them in a batch computation.
I would really raise an error, and maybe, let the user consciously turn on a parameter (USE_OUT_OF_RANGE_PARTITION_FUNCTION)  to allow this behavior.

Then it's the users's problem and not the fault of the exojax developers ;)

erwanp (changes requested):

ykawashima:
Thank you for your suggestion and for sharing the plot. I understand. Then shall we prepare such flag for the moldb class? @HajimeKawahara What do you think?

HajimeKawahara (commented):
Thank you for implementing setdgm for Hitemp! 
I think it is useful to include some test or simple example code for MODIT/HITEMP, for instance, a HITEMP version of `exojax/tests/manualtest/reverse/reverse_methane.py` (this example uses exomol for CH4 retrieval).  Can you make such an example?

HajimeKawahara:
> Then shall we prepare such flag for the moldb class? @HajimeKawahara What do you think?

How about use 

```python
raise ValueError("Error: partition function ratio for isotope ... ")
```
instead of `print`?

ykawashima:
> I think it is useful to include some test or simple example code for MODIT/HITEMP, for instance, a HITEMP version of exojax/tests/manualtest/reverse/reverse_methane.py (this example uses exomol for CH4 retrieval). Can you make such an example?
OK, I will do that tomorrow.

So you mean removing the option to calculate for the temperatures outside the supported range?
This maybe just for my case, but some molecules such as H2S have relatively lower Tmax and I need to do the calculation for the temperature just 10 or 20 K higher than that. So for me, it would be great to have the option to do the calculation outside the supported range. Of course, I agrree that in that case, the users should know that the result maybe wrong.

HajimeKawahara:
>OK, I will do that tomorrow.
Thanks a lot in advance.

>I need to do the calculation for the temperature just 10 or 20 K higher than that

I see. Then shall we leave `print` for now and think about flag or error in  a forthcoming feature?

erwanp:
@ykawashima if it's really close to the supported range, should be fine. You could add some quick (but not too dirty) solutions : 

1.  check that temperature is only <1% outside of the range; clip if it is ; raise an error if it is not 
2.  extrapolate : partition functions can be fitted by a 2 or 3 order polynomial (look at the calculation above); if outside the range you could fit an order 2 polynomial on the tabulated HAPI data; and extrapolate.  Again, raise an error if <20% outside the temperature range

ykawashima:
@erwanp Thank you for the suggestion. I agree it would be great to implement a polynomial fit.

I considered this matter and have changed my mind (sorry). I now think that it is better not to officially provide the calculation outside the supported temperature range as it is a bit dangerous as @erwanp pointed out. So I will use the clip just for my environment and implement the polynomial fit in the future official release. Is that fine for both of you?

erwanp:
Yes, very safe!

ykawashima:
While creating a test retrieval code for HITRAN/HITEMP, I realized that we first need to make [Qr_layer_HAPI](https://github.com/HajimeKawahara/exojax/blob/3799248cff527ff475adf6ac9d2f57ba2ebd683e/src/exojax/spec/moldb.py#L810) capable of JAX. So I will close this pull request for now and reopen it after implementing that. Sorry for the confusion. I really appreciate your suggestions!!

PR #246, opened 2022-04-14
Hitemp rename

ykawashima:
Sorry for the multiple pull requests, but I have added a feature to rename the HITEMP path when ".bz2" is missing. (I thought it is a bit confusing as HITEMP requires ".par.bz2" path while HITRAN requires ".par" path.) Thank you in advance.


HajimeKawahara (approved):
Thanks!

PR #250, opened 2022-04-25
Kurucz

chonma0ctopus:
The support for the Kurucz linelist was confirmed with minor changes and an example notebook is added.
(In addition, I also confirmed support for the "Extract Stellar" mode of the VALD3.)


HajimeKawahara (approved):
Thanks! Confirmed that the test worked.

PR #253, opened 2022-05-08
Setdgm HITRAN/HITEMP

ykawashima:
Sorry for taking the time to implement this (I am unfamiliar with JAX), but I have modified the code to use setdgm also for HITRAN/HITEMP database by changing the calculation part of the partition function from HAPI to [jnp.interp](https://github.com/ykawashima/exojax/blob/e6af6ddd8bd51e3c700a045e18790990a1bc7edb/src/exojax/spec/moldb.py#L723) with the HAPI data. You can test this with [exojax/tests/manualtest/reverse/reverse_CO_HITEMP.py](https://github.com/ykawashima/exojax/blob/e6af6ddd8bd51e3c700a045e18790990a1bc7edb/tests/manualtest/reverse/reverse_CO_HITEMP.py).

There are three things I would like to note and ask you.

1. DIT grid matrix is prepared with the constant reference partial pressure array (see [here](https://github.com/ykawashima/exojax/blob/e6af6ddd8bd51e3c700a045e18790990a1bc7edb/src/exojax/spec/modit.py#L329)) given that self broadening is relatively small. Are you fine with this or do you think it is better to have the axis also for Pself?
2. As the mask feature (like [this line](https://github.com/ykawashima/exojax/blob/e6af6ddd8bd51e3c700a045e18790990a1bc7edb/src/exojax/spec/moldb.py#L790)) does not seem to be allowed in the JAX environment, I used for loop for all the self.isoid in Qr_line_HAPI_jax (see [here](https://github.com/ykawashima/exojax/blob/e6af6ddd8bd51e3c700a045e18790990a1bc7edb/src/exojax/spec/moldb.py#L810)). Do you have any better solution?
3. While I prepared an error comment (see [here](https://github.com/ykawashima/exojax/blob/e6af6ddd8bd51e3c700a045e18790990a1bc7edb/src/exojax/spec/moldb.py#L763-L766)) when the temperature is outside the supported range of Qr_HAPI, I could not prepared such feature for Qr_line_HAPI_jax given that print statement is not allowed in the JAX environment. For this point too, do you have any idea?

Thank you in advance!


HajimeKawahara:
Thanks! @ykawashima 
When running `reverse_CO_HITEMP.py,` I needed to manually make the directory:
```sh
mkdir .database/CO/05_HITEMP2019
```
but after this, it worked.

ykawashima:
Thank you. Yes, the auto-directory-creating feature is not available for some HITRAN/HITEMP cases as far as I remember. I will have a look.

ykawashima:
I added one line in modlb.py to automatically create a directory for HITRAN/HITEMP.

What do you think about the following three things?

> 1. DIT grid matrix is prepared with the constant reference partial pressure array (see [here](https://github.com/ykawashima/exojax/blob/e6af6ddd8bd51e3c700a045e18790990a1bc7edb/src/exojax/spec/modit.py#L329)) given that self broadening is relatively small. Are you fine with this or do you think it is better to have the axis also for Pself?
> 2. As the mask feature (like [this line](https://github.com/ykawashima/exojax/blob/e6af6ddd8bd51e3c700a045e18790990a1bc7edb/src/exojax/spec/moldb.py#L790)) does not seem to be allowed in the JAX environment, I used for loop for all the self.isoid in Qr_line_HAPI_jax (see [here](https://github.com/ykawashima/exojax/blob/e6af6ddd8bd51e3c700a045e18790990a1bc7edb/src/exojax/spec/moldb.py#L810)). Do you have any better solution?
> 3. While I prepared an error comment (see [here](https://github.com/ykawashima/exojax/blob/e6af6ddd8bd51e3c700a045e18790990a1bc7edb/src/exojax/spec/moldb.py#L763-L766)) when the temperature is outside the supported range of Qr_HAPI, I could not prepared such feature for Qr_line_HAPI_jax given that print statement is not allowed in the JAX environment. For this point too, do you have any idea?

HajimeKawahara:
Thanks! I confirmed that `.database/CO/05_HITEMP2019` was automatically created!

HajimeKawahara (approved):
I approve this PR, but if you want to continue to improve the code in terms of the three points you raised, let me know. I will wait until then.

ykawashima:
Thank you for your comments. That's true that we should keep that difference in mind.

I would like to implement the mask feature for now while I prefer to leave the other two points for future implementations. Could you wait a bit? Thank you.

HajimeKawahara:
yes, agreed. Thanks.

ykawashima:
I have finished modifying the code. Please merge the PR if you are fine with it. Thank you.

HajimeKawahara:
Thanks! I approve the PR.

PR #267, opened 2022-08-04
added a warning when no lines are selected.

ykawashima:
I just added a warning when the molecular database contains no lines as it sometimes causes an error which is hard to understand the problem.


HajimeKawahara:
Can you use 

``` python
import warnings
warnings.warn("message", UserWarning)
```
for warning messages?

like [here](https://github.com/HajimeKawahara/exojax/blob/ed760cc8720e91edee6d2aaf16fb3180873f4c17/src/exojax/spec/modit.py#L165).

ykawashima:
Thank you. I have changed.

PR #269, opened 2022-08-18
HITRAN molmass

ykawashima:
I have added an option to use the molecular mass provided by the HITRAN database;
https://hitran.org/docs/iso-meta/

Thank you in advance.


HajimeKawahara:
Thanks! Can you add a short unit test?
for instance to check

```python
molmass('CO2',db_HIT=True)
```

ykawashima:
I have added tests/unittests/atom/molmass_test.py. This is the first time for me to write a unit test. Is the code fine with you?

tests/unittests/atom/molmass_test.py
```diff
@@ -0,0 +1,16 @@
+"""test for molmass calculation"""
+
+import pytest
+from exojax.spec.molinfo import molmass
+
+
+def test_molmass():
+    molmass('air')
```
erwanp, line 8: (if I can suggest, you should add `assert` statements there
ex: 
```python
assert molmass('air')== 28.97
```

erwanp: They'll raise an AssertionError if it doesn't work. 

Advanced note : sometimes (not here) you want to test that an error is raised, i.e., the normal behavior is to expect an error (for instance, you're trying to catch a wrong user input, etc.) 
You can use 
```python 
with pytest.raises(ValueError):
    this_code()   #  test will pass IF this_code() raises a ValueError
```

ykawashima: Thank you very much for your suggestion and helpful tip! I will implement it.

HajimeKawahara (approved):
Thank you! All of the unit tests passed.

Issue #277, opened 2022-08-31
Aborted (core dumped) in FFT with FP64


chonma0ctopus:
For applying HMC fitting to observed spectra, I employed FP64 to avoid the divergences reported in #268. 
I encountered the following. It looks related to google/jax#9591 and tensorflow/tensorflow@e75ac1511159b10b1a2c29291d21fd3407eb703b.

```
2022-08-28 12:49:18.328674: E external/org_tensorflow/tensorflow/stream_executor/cuda/cuda_fft.cc:253] failed to allocate work area.
2022-08-28 12:49:18.328712: E external/org_tensorflow/tensorflow/stream_executor/cuda/cuda_fft.cc:430] Initialize Params: rank: 1 elem_count: 400 input_embed: 400 input_stride: 1 input_distance: 400 output_embed: 201 output_stride: 1 output_distance: 201 batch_count: 1336566
2022-08-28 12:49:18.328719: F external/org_tensorflow/tensorflow/stream_executor/cuda/cuda_fft.cc:439] failed to initialize batched cufft plan with customized allocator: Allocating 4298396256 bytes exceeds the memory limit of 4294967296 bytes.
Aborted (core dumped)
```

---------------------------------------------------------------------------------

The full input and output are below.
Sorry for the too-long code that may include parts not be relevant. I'm hoping to break down the issue when I have time, but I just wanted to let you know the situation. (If you want to reproduce the situation, I can send you the data used in the code.)

[Input]: 
```python
path_fig = '/home/tako/ghR/ghR/cfca/tmpfig/fit_GJ699_woFastChem_TPlbl/'
Nth_run = "test220827atR740"
import os
os.environ['XLA_PYTHON_CLIENT_PREALLOCATE'] ='false'
os.environ['XLA_PYTHON_CLIENT_ALLOCATOR']='platform'
os.environ['TF_FORCE_GPU_ALLOW_GROWTH'] = 'true'
from jax.config import config
config.update("jax_enable_x64", True)


path_ObsData = '/home/tako/ghR/ghR/cfca/data/gj699_coadded_cr2.dat'
path_ValdLineList = '/home/tako/ghR/ghR/cfca/data/HiroyukiIshikawa.4216193.gz'

import arviz
from cuda import cudart
import numpy as np
import matplotlib.pyplot as plt
import jax.numpy as jnp

from exojax.spec.rtransfer import pressure_layer
from exojax.spec.setrt import gen_wavenumber_grid
[...]

PR #285, opened 2022-09-26
Modit scanfft

HajimeKawahara:
This PR addresses #277.
Instead of replacing MODIT core algorithm, I just added the scan+fft version of MODIT. The reasons are

- standard core algorithm for large lines will be replaced to `premodit` based approach, presumably `presolar`
- The fft 4GB limit presumably will be resolved in future.
- The current version of `modit_scanfft` is 2 times slower than the original `modit`


chonma0ctopus (commented):
Thank you! Confirmed that all unit tests passed. 

However, the error on my sample code in #277 needs the _scanfft version of `xsmatrix_vald_scanfft`.
I add it and commit here. Is it ok?

HajimeKawahara:
Thanks!

> However, the error on my sample code in https://github.com/HajimeKawahara/exojax/issues/277 needs the _scanfft version of xsmatrix_vald_scanfft.
> I add it and commit here. Is it ok?

Yeah, sure. Go ahead.

PR #286, opened 2022-09-27
Modit scanfft vald

chonma0ctopus:
Added the scanfft version of `xsmatrix_vald_scanfft` (as same as the functions for exomol database in #285) and the corresponding testfiles.


HajimeKawahara:
I had an assertion error in `modit_scanfft.py::test_rt_vald` 

```sh
        print(np.max(residual))
>       assert np.all(residual < 0.03)
E       assert False
E        +  where False = <function all at 0x7f0b27da6040>(array([nan, nan, nan, ..., nan, nan, nan]) < 0.03)
E        +    where <function all at 0x7f0b27da6040> = np.all

spec/modit_scanfft/modit_scanfft_test.py:158: AssertionError
------------------------------------------- Captured stdout call --------------------------------------------
xsmode assumes ESLOG: mode= modit
nan
```
The residual was NaN.

HajimeKawahara:
I found that this was case dependent. It passes sometimes but fails other time.

HajimeKawahara:
Adding 
```python
from jax.config import config
config.update("jax_enable_x64", True)
```
in `modit_scanfft.py`,
The error became reproducible (always NaN error)

Can you check this? @chonma0ctopus

chonma0ctopus:
Thank you for pointing this out! You are right.
The fundamental cause of this problem has been found in `modit.set_ditgrid_matrix_vald_each`. I added one line as a workaround there.

HajimeKawahara (approved):
Thanks. All of the unittests passed.

PR #301, opened 2022-10-25
Extract tau=1 height pressure

chonma0ctopus:
Added a module to extract the pressure at the height of tau = 1 (or another given value) at different wavelengths, and its unittest and example notebook.
Note that the module will return 0 if tau is too optically thin and does not reach tau=1.


src/exojax/spec/tau1height.py
```diff
@@ -0,0 +1,24 @@
+"""Extract tau=1 height pressure at different wavelengths"""
+import numpy as np
+from scipy import interpolate
+
+def get_P_at_tauone(dtau, Parr, tauextracted=1.0):
```
HajimeKawahara, line 5: `get_P_at_tauone` -> `extract_pressure_at_tauone`? (from the comment)
or simply `pressure_at_given_opacity`? any good idea?

chonma0ctopus: Thanks for the suggestion! Okay, I'll choose `pressure_at_given_opacity`.

src/exojax/spec/tau1height.py
```diff
@@ -0,0 +1,24 @@
+"""Extract tau=1 height pressure at different wavelengths"""
```
HajimeKawahara, line 1: Not sure but, looks a bit rare to include "1" in the name of modules. I have no good idea but for instance
- tauinfo (assuming we include more analysis functions in it)
- opacitytools 
-  ... no good idea...

HajimeKawahara: i mean i talk about "tau1height.py" just in case

HajimeKawahara:
Thanks! I had some comments for names.

src/exojax/spec/tau1height.py
chonma0ctopus: Thanks! I'll adopt `opacitytools`.

HajimeKawahara:
@chonma0ctopus 
What is `modit_issue277.py`?

chonma0ctopus:
> @chonma0ctopus What is `modit_issue277.py`?

Sorry, I mis-pushed... never mind please.

chonma0ctopus:
Now, it is fixed and all the suggested name changes have been done.

HajimeKawahara (approved):
Thanks! Cool figure!
image

PR #303, opened 2022-10-26
bugfix for HITEMP/HITRAN HMC

HajimeKawahara:
When implementing `api`, I made a bug in `MdbHitemp` and `MdbHitran`:
I used `jnp array` for `self.uniqiso`, but this breaks JITtable feature. 
I restored them to `np.array`.

I rewrote `exojax/tests/endtoend/reverse/reverse_modit_hitemp.py` using `api`.


HajimeKawahara:
Related to #253

ykawashima (approved):
Thank you very much. I have confirmed that exojax/tests/endtoend/reverse/reverse_modit_hitemp.py works.

HajimeKawahara:
Thanks!

PR #309, opened 2022-11-07
replaced moldb.MdbExomol and MdbHit to api.MdbExomol and MdbHit

HajimeKawahara:
Deleted `moldb.MdbExomol `and `Mdb.Hit`.
Meanwhile I put the aliases to `api.Mdb***` to keep compatibility, but we will remove these aliases when releasing `ExoJax2 `


chonma0ctopus (approved):
Thank you very much!

PR #312, opened 2022-11-08
tests prior to the release process

HajimeKawahara:


HajimeKawahara:
[The release 0.14](https://github.com/radis/radis/pull/532) of `radis` solves some fails in the unit test. Now, there is no fails in the github-action!

Issue #314, opened 2022-11-09
np.unique RAM issue 


HajimeKawahara:
@ykawashima pointed out that numerous RAM is used in `init_premodit` in the case of our favorite CH4 ExoMol database. This looks actually from [here](https://github.com/HajimeKawahara/exojax/blob/33a82c308c50740f8b50a1be6436d8ffd0fce3c9/src/exojax/utils/indexing.py#L21).

```
uniqvals = np.unique(input_array, axis=0)
```

I found there is some info about `np.unique(a,axis=0)` on [here](https://github.com/numpy/numpy/issues/15713). I will see.

HajimeKawahara:
According to [this post](https://discuss.dizzycoding.com/find-unique-rows-in-numpy-array/?amp=1), Answer #4, 

```python
def unique_rows(data):
    uniq = np.unique(data.view(data.dtype.descr * data.shape[1]))
    return uniq.view(data.dtype).reshape(-1, data.shape[1])
```
looks save RAM.

HajimeKawahara:
But, this makes such error... for large memory
```python
E       ValueError: When changing to a larger dtype, its size must be a divisor of the total size in bytes of the last axis of the array.
```

erwanp:
Can you do premodit fully in Vaex? This would fix your RAM problem

HajimeKawahara:
Ah, good idea! I will consider this possibility.

HajimeKawahara:
Meanwhile, #315 implemented the above code.  @ykawashima confirmed that it improved the RAM use when CH4/ExoMol is used.

PR #315, opened 2022-11-10
Initspec memory issue

HajimeKawahara:
This PR is to address #314.
Can you check if it works or not in your case? @ykawashima 
(because I do not have a sample (and smaller RAM device) that ends up a RAM overflow)


ykawashima (approved):
Thank you very much for the quick implementation. I have confirmed that the same code now works without yielding the memory shortage error.

PR #320, opened 2022-11-20
Solved the problem of too long times taken by PLG with the latest JAX.

chonma0ctopus:
When I updated JAX to 0.3.24 from 0.2.22, the time taken for PLG was increasing strangely. The cause was a situation where jax.numpy.array was included in the args of python-build-in min/max functions. I replaced the min/max with numpy.min/numpy.max.


HajimeKawahara:
Oh, really? Good to know. Thanks for finding that changes.

HajimeKawahara (approved):

PR #328, opened 2022-12-23
Molmass isotope

HajimeKawahara:
So far, the mean molecular mass has been used as MolMass. I newly made an isotope-by-isotope based one. 

- MdbExomol has a new instance of .molmass 
- MdbHitemp not yet.


src/exojax/spec/molinfo.py
```diff
@@ -27,63 +42,182 @@ def molmass(molecule, db_HIT=False):
...
-
-
-    if molecule == 'air' or molecule == 'Air':
+    if simple_molecule_name == 'air' or simple_molecule_name == 'Air':
         return 28.97
 
-    if molecule in molmass_hi and db_HIT:
-        molmass = molmass_hi[molecule]
+    if simple_molecule_name in molmass_isotope and db_HIT:
+        molmass = molmass_isotope[simple_molecule_name][0]
```
ykawashima, line 51: If you want to extract the mass for the most abundant isotope combination, I think this line should be
```suggestion
        molmass = molmass_isotope[simple_molecule_name][1]
```
since [0] corresponds to the mean molar mass as written [here](https://github.com/HajimeKawahara/exojax/blob/e1745de3736c1c469b58297f4c7ca56561fbb12f/src/exojax/utils/isotopes.py#L82).
I suppose this is why I currently get different results from `molinfo.molmass_isotope("CO")` and `molinfo.isotope_molmass("12C-16O")`.

src/exojax/utils/isotopes.py
```diff
@@ -1,17 +1,118 @@
...
+    Returns:
+        dict:  molmass_isotope, abundance_isotope
+
+    Examples:
+
+        >>> path = pkgutil.get_data('exojax', 'data/atom/HITRAN_molparam.txt')
+        >>> mean_molmass, molmass_isotope, abundance_isotope = read_HITRAN_molparam(path)
+        >>> molmass_isotope["CO"][1] # molar mass for CO isotope number = 1
+        >>> abundance_isotope["CO"][1] # relative abundance for CO isotope number = 1
+        >>> molmass_isotope["CO"][0] # mean molar mass for CO isotope number = 1
```
ykawashima, line 82:
```suggestion
        >>> molmass_isotope["CO"][0] # mean molar mass for CO
```

ykawashima (commented):
Thank you very much. I have added two suggestions.

Also, as far as I understand, [this line](https://github.com/HajimeKawahara/exojax/blob/e1745de3736c1c469b58297f4c7ca56561fbb12f/src/exojax/spec/molinfo.py#L29) should be 
`db_HIT: if True, use the molecular mass provided by HITRAN (https://hitran.org/docs/iso-meta/)`

src/exojax/spec/molinfo.py
```diff
@@ -1,13 +1,28 @@
...
+from exojax.utils.molname import e2s
+from exojax.utils.isotopes import exact_molecule_name_to_isotope_number
 
+def isotope_molmass(exact_molecule_name):
+    """isotope molecular mass 
 
-def molmass(molecule, db_HIT=False):
-    """provide molecular mass from the molecular name.
+    Args:
+        exact_molecule_name (str): exact exomol, hitran, molecule name such as 12C-16O,  (12C)(16O)
```
ykawashima, line 10: In my environment, `molinfo.isotope_molmass("(12C)(16O)")` does not work, while `molinfo.isotope_molmass("12C-16O")` does. Is that correct?

HajimeKawahara: Thanks! I fixed this bug.

Issue #337, opened 2023-01-27
HITEMP/HITRAN quantum states filtering


HajimeKawahara:
Quantum states filtering was implemented for ExoMol #336 but not yet HITEMP and HITRAN.
It looks the original file contains such information:

 - "global_upper_quanta": "Electronic and vibrational quantum numbers and 
 -  "global_lower_quanta": "Electronic and vibrational quantum numbers and labels for the lower state of a transition"

HajimeKawahara:
@YuiKasagi

erwanp:
Its implemented in Radis for CO2 and CO (therefore in the api)

erwanp:
Parsing was also done for many hitran molecules  https://github.com/radis/radis/pull/441

HajimeKawahara:
Oh, cool! I'll try it.

erwanp:
Just check that it's what you need in Exojax. We can discuss on Slack

PR #338, opened 2023-01-28
quantum state filtering for  hitran/hitemp

HajimeKawahara:
I made the HITEMP and HITRAN version of quantum state filter. #337 

See 

- tests/integration/moldb/quantum_states_filter_hitemp.py  
- tests/integration/moldb/quantum_states_filter_hitran_co.py
- tests/integration/moldb/quantum_states_filter_exomol.py

But, I found that O2 in HITRAN does not contain the information on vibration states....


YuiKasagi (approved):
Thank you! I confirmed that all figures in the new test codes could be drawn and looked good. (I understand the situation with O2 in HITRAN...)

HajimeKawahara:
Thanks!

Issue #356, opened 2023-03-15
transmission spectroscopy


HajimeKawahara:
Started to implement. Still wrong though...

HajimeKawahara:
![comparison](https://user-images.githubusercontent.com/15956904/225185669-cb7b8cd3-3ec5-4bca-9699-0c4332958d2e.png)
something wrong...

HajimeKawahara:
![Figure_1](https://user-images.githubusercontent.com/15956904/225248177-db39f48f-99ba-4f29-bf1d-807711be5fae.png)

HajimeKawahara:
getting close...

HajimeKawahara:
The difference is likely due to the lack of Rayleigh scattering.

HajimeKawahara:
closed #359

PR #359, opened 2023-03-18
Frist implementation of Transmission Spectrum Computation

HajimeKawahara:
This is a first attempt to compute a transmission spectrum of CO #356 , and a comparison with @ykawashima data.

Can you confirm the results of the following code?

exojax/tests/integration/comparison/transmission/comparison_with_kawashima_transmission.py

I believe the difference between two is due to the lack of Rayleigh scattering, which I plan to implement next. Thanks in advance.

comparison_with_kawashima_transmission


astrostevanus:
Thanks for the PR. I have tried the example script that you provided (comparison_with_kawashima_transmission.py) and it works without any problem. Then, I tried to compare with petitRadTrans output (pure CO, no neither rayleigh scattering nor CIA, same parameters as in Exojax's script) although using a different line list (all isotope, HITEMP instead of HITRAN). Overall, the transit radius seems to agree, but the big difference is in the longer wavelength, I am not sure if it is due to different line wing cut-off limit that was used or other possible causes.

comparison_general

HajimeKawahara:
thanks! Do you know which broadening parameters petitRadtrans uses for HITEMP?
Both Kawashima-san and I are using the self-broadening because it's 100% CO.

astrostevanus:
No worries! Unfortunately, they do not include that information in the website. I found in the paper that they include pressure broadening by either using the air broadening coefficients of HITRAN/HITEMP (Rothman et al.2010, 2013), or Eq. (15) of Sharp & Burrows (2007).

ykawashima (approved):
Thank you very much for the implementation! I have confirmed that the code works without any problem in my environment, too.

astrostevanus (approved):

HajimeKawahara:
The bug fix #414 improved the agreement with Kawashima-san's results.
Figure_1

PR #369, opened 2023-04-03
Bug fix for pressure layer

ykawashima:
This pull request is to fix a bug for `pressure_layer`, which I just found out. Due to the lack of the "reference_point" parameter in `pressure_layer_logspace`, "numpy" is regarded as the value for "reference_point",  and this results in a zero array for dParr in the current develop branch. You can check the update with the following two lines. Thank you in advance.

```
from exojax.spec.rtransfer import pressure_layer
Parr, dParr, k=pressure_layer(NP=100)
print(dParr)
```


HajimeKawahara (approved):
Thanks!

Issue #366, opened 2023-03-21
grid optimization for broadening parameters


HajimeKawahara:
I postponed this issue when releasing v1.3. 
Related to #365,  we realized that the appropriate grid size of broadening parameters was important for real life ;)

- should be considered with #327.

HajimeKawahara:
As discussed with @ykawashima, we decided to make linear grids of broadening parameters instead of log.

HajimeKawahara:
Memo: Two equations to think about the errors of the current implementation

![image](https://user-images.githubusercontent.com/15956904/229380192-eb60f8f4-a05b-4248-b18e-a99c7978b4f5.png)

![image](https://user-images.githubusercontent.com/15956904/229380228-61370c22-a6c3-4586-a18c-3ec4ff79f749.png)

A potential improvement is we change `Tref` for n. For instance, when our range is 400-2000K, the current implementation requires, np.log(2000.)-np.log(273.) = 2 of the boost factor. but if setting Tref at the middle of log 400 and 2000 K, 

```python
>>> tmid=(np.log(2000.)+np.log(400.))/2
>>> np.log(2000.)-tmid
0.8047189562170498
>>> (np.log(2000.)-np.log(273.))/(np.log(2000.)-tmid)
2.474690883036675
```
we get about 2.5 times smaller boost factor, which reduces the device memory use.

HajimeKawahara:
The 2D C should be 4/3 times smaller than the 1D case:

![diff2D1D_broadening](https://user-images.githubusercontent.com/15956904/229382931-4de419e1-cd3f-4d61-b94b-cadbd8f73b43.png)

HajimeKawahara:
example: `test_broadpar_grid_as_a_function_of_Tref_broadening(db)` in `tests/unittests/spec/premodit/premodit_test.py`

HajimeKawahara:
N_gamma, N_n

|     |current |this  |
|-----|---------|------|
| H2O | 21, 17|28, 8|
| CH4 | 5, 13  |12, 7|
| CO|5, 2     |5, 2  |
| NH3|8, 6     |13, 3|
| H2S|9, 2     |12, 2|

HajimeKawahara:
implemented #370

Issue #327, opened 2022-12-20
Negative self.n_air for HITRAN/HITEMP


ykawashima:
It seems that some HITRAN/HITEMP database (e.g., HITRAN/HITEMP H2O) contains negative self.n_air values. Should they indeed be positive as written [here](https://github.com/HajimeKawahara/exojax/blob/c6fbc2e787285b4d508a3e2416e5bb5c8f9b931c/src/exojax/spec/set_ditgrid.py#L51)? Thank you.

erwanp:
They should be, it's for sure a glitch during the generation of the database. I'd either discard all lines or take the n_self value (if given), and report to the HITRAN team

ykawashima:
Thank you very much for your comment. I understand. Then, I will discard those lines.

HajimeKawahara:
@erwanp Do you know why the temperature exponent (n_air) should be positive?

erwanp:
My bad, the gamma_air should obviously be positive but n_air shouldn't. I'll have a second look

HajimeKawahara:
I see, thanks. So, this issue should be tagged as "bug".

HajimeKawahara:
Just note: slight deviation between PreMODIT and MODIT originates this bug?

HajimeKawahara:
Maybe I encountered this error. When using HITRAN and MODIT for CO. 

```
   opa = OpaModit(mdb=mdb, nu_grid=nu_grid, Tarr_list=Tarr, Parr=np.zeros_like(art.pressure), Pself_ref = art.pressure)
  File "/home/kawahara/exojax/src/exojax/spec/opacalc.py", line 266, in __init__
    self.setdgm(Tarr_list, Parr, Pself_ref=Pself_ref)
  File "/home/kawahara/exojax/src/exojax/spec/opacalc.py", line 352, in setdgm
    dgm_ngammaL = precompute_modit_ditgrid_matrix(
  File "/home/kawahara/exojax/src/exojax/spec/set_ditgrid.py", line 145, in precompute_modit_ditgrid_matrix
    grid = np.logspace(lxmin, lxmax, Ng)
  File "<__array_function__ internals>", line 5, in logspace
  File "/home/kawahara/anaconda3/lib/python3.8/site-packages/numpy/core/function_base.py", line 275, in logspace
    y = linspace(start, stop, num=num, endpoint=endpoint, axis=axis)
  File "<__array_function__ internals>", line 5, in linspace
[...]

PR #370, opened 2023-04-04
grid broadening optimization

HajimeKawahara:
This PR implemented #366

- removed assertion for the negative n #327 
- re-parametrize gamma according to updated reference temperature for the broadening (i.e. `Tref_broadening`).
- OpaPremodit assumes the log midpoint of temperature for `Tref_broadening`
- this reduces the number of grid for temperature exponent and therefore device memory use, given accuracy (C_gamma in DIT paper)
- we use more (x 3/4) strict value as C_gamma as described in #366, which increases the device memory use.
- in total about it's 1 - 1.6 times lesser device memory efficient, depending on the database. 

See #366 for some experiments.


ykawashima (approved):
Thank you very much! I have confirmed the result of ``test_broadpar_grid_as_a_function_of_Tref_broadening(db)`` in ``tests/unittests/spec/premodit/premodit_test.py``.

PR #376, opened 2023-04-10
Single broadening parameter

HajimeKawahara:
I made the option that assume a single broadening parameter set in PreMODIT.
For unit tests, see 
- tests/unittests/spec/premodit/premodit_spectrum_test.py 
- tests/unittests/spec/premodit/premodit_xsection_test.py 

I also edit [the document](http://secondearths.sakura.ne.jp/exojax/develop/userguide/premodit.html#single-broadening-parameter-set). Check it.


ykawashima (approved):
I apologize for the delay in reviewing. I have confirmed that the unit tests you mentioned work. Thank you very much.

Issue #382, opened 2023-05-13
max-min grid for broadening parameter and refactoring dit_grid_resolution in OpaPremodit


HajimeKawahara:
@ykawashima found that the minimum set of broadening parameters (i.e. using max and min values of temperature exponent and gamma_air) was sometimes enough for real spectra. The difference between `dit_grid_resolution=0.1` and the max-min case was within 1%. So, I would make this option in addition to `single_broadening` option #376 . 

This reduces the device memory use a lot.

Also, I would refactor these options...

PR #383, opened 2023-05-14
Broadpar minmax mode and plot of values and grids  OpaPremodit

HajimeKawahara:
Addressed #382.

See https://secondearths.sakura.ne.jp/exojax/develop/userguide/premodit.html for the details
The test is given in `tests/integration/broadening/broadpar_hitemp.py`

- `single_broadening` option in `OpaPremodit` was removed
- instead, we use `broadening_resolution` option, which enables us to switch "manual", "minmax", and "single" modes.

- `opa.plot_broadening_parameters()` can make a plot such as 

example_manual


ykawashima (approved):
Thank you very much! I have confirmed that the test code works for the three modes.

PR #386, opened 2023-05-20
Fixed error for float values in MdbExomol and added the use of broadf

YuiKasagi:
I fixed minor bugs in MdbExomol.
- Convert values of jlower / jupper to integers: compute_broadening() only accepts integers, but I found SO2 has float values.
-  Added the use of the option of `broadf`: If this is False, the default broadening parameters in .def file are used in radis. We can avoid downloading the .def file with this option after I fix a code in radis. (e.g. There is no .def file for OH.)


HajimeKawahara (approved):
All of the unit tests passed. Thanks!

PR #387, opened 2023-05-31
removed negative value assert (hotfix to v1.3)

HajimeKawahara:
We found a bug, which affects HITEMP/H2O opacity.
- in set_grid.py the negative value was not allowed for the density grid. So, users could have removed the negative temperature exponent of the HITEMP/H2O. But, this modification could have removed some important H2O lines.

the following is the comparison, @ykawashima made: The blue lines removed the lines with n_air < 0.01, yellow clipped to n_air = 0, green no restriction for HITEMP H2O. 
image


ykawashima (approved):
Thank you very much! I have confirmed that the code can now treat the lines with negative n_air.

PR #390, opened 2023-06-06
test setup.py

HajimeKawahara:
I'd like to check CI status.


HajimeKawahara:
See https://github.com/actions/setup-python/issues/160

Issue #391, opened 2023-06-07
wavenumber and wavelength descending/ascending order


HajimeKawahara:
Currently, `wavenumber_grid` returns `nu_grid` and `wav` both in ascending order. So, for instance, we should plot
```python
plt.plot(nu_grid, flux)
plt.plot(wav[::-1], flux)
```
this might be confusing. rethink user-friendly (how?) and developer-friendly (wavenumber in ascending order) definition.

PR #395, opened 2023-06-13
Wavelength order

HajimeKawahara:
addressed #391 

tentative new policy on the order of wavenumber/wavelength
- wavenumber grid should be in ascending
- users can choose the order of wavelength grids via `wavelength_order` in `utils.grids.wavenumber_grid` or `spec/unitconvert/nu2wav` but the default order is **descending** (the old one assumed **ascending**)


ykawashima (approved):
Thank you very much! I have confirmed that the wavelength_order option for wavenumber_grid, nu2wav, and wav2nu all work.

PR #399, opened 2023-06-16
Release version1.4

HajimeKawahara:
We started the release process of [version 1.4](https://github.com/HajimeKawahara/exojax/milestone/7).
Unless a severe bug is found, we plan to release v1.4 by July 14th (a bit far from now because we need time after ExoClimes).  

I will write more documents in this branch.


src/exojax/spec/opacont.py
```diff
@@ -27,11 +27,14 @@ class OpaCIA(OpaCont):
...
     """
-    def __init__(self, cdb, nu_grid):
+    def __init__(self, cdb, nu_grid, wavelength_order="descending"):
         self.method = "cia"
         self.warning = True
         self.nu_grid = nu_grid
-        self.wav = nu2wav(self.nu_grid, unit="AA")
+        self.wavelength_order = wavelength_order
+        self.wav = nu2wav(self.nu_grid,
+                          wavelenght_order=self.wavelength_order,
```
YuiKasagi, line 36: I found that there was a typo "waveleng'ht'_order".

HajimeKawahara: Thanks!

src/exojax/spec/multimol.py
```diff
@@ -0,0 +1,261 @@
...
+    Returns:
+        str: Exomol default data path
+    """
+    from exojax.utils.molname import simple_molname_to_exact_exomol_stable
+    from radis.api.exomolapi import get_exomol_database_list
+    exact_molname_exomol_stable = simple_molname_to_exact_exomol_stable(
+        simple_molecule_name)
+    mlist, recommended = get_exomol_database_list(simple_molecule_name,
+                                                  exact_molname_exomol_stable)
+    dbpath = database_root_path + "/" + simple_molecule_name + "/" + recommended
```
YuiKasagi, line 260: This line caused a problem in my analysis. I think this should be the following.:
`dbpath = simple_molecule_name + "/" +  exact_molname_exomol_stable + "/" + recommended`

(For example, it reterns `./database/H2O/POKAZATEL` now, but it failed to download database. `H2O/1H2-16O/POKAZATEL` is correct.)

And we do not need `database_root_path` option for this function, like `database_path_hitemp` and `hitran`.

HajimeKawahara: Thanks! After correcting this line, did it work?

YuiKasagi: Yes, it worked.

HajimeKawahara: corrected. Can you check if it works?

YuiKasagi: Thank you! I confirmed that it works correctly.

PR #409, opened 2023-08-05
doc update (to describe the recommended installation to remove older modules)

HajimeKawahara:
See https://secondearths.sakura.ne.jp/exojax/userguide/installation.html

### DIFF
--- a/documents/userguide/installation.rst
+++ b/documents/userguide/installation.rst
@@ -19,7 +19,7 @@ Alternatively, clone the code from `github page <https://github.com/HajimeKawaha
 	  
 	  python setup.py install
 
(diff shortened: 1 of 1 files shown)


documents/userguide/installation.rst
```diff
@@ -19,6 +19,13 @@ Alternatively, clone the code from `github page <https://github.com/HajimeKawaha
 	  
 	  python setup.py install
 
+If the older version of ExoJAX has already been installed, you need to remove all of the old modules. To do so, the following procedure is recommended:
```
ykawashima, line 22:
```suggestion
If the older version of ExoJAX has already been installed, you need to remove all of the old modules (Otherwise, modules that have been deleted and no more exist in the current version remain). To do so, the following procedure is recommended:
```

ykawashima (approved):
Thank you very much!

Issue #241, opened 2022-03-21
radiative transfer rethinking


HajimeKawahara:
In Paper I, we used 2 E_3 as a pure absorption transmission T, but not tried a numerical integral of mu e^-dtau/mu, for instance, using Gaussian quadrature etc, which might reduce a computational cost than the direct computation of 2 E_3. Even using mu=2/3,  e^-dtau/mu,  is not a bad approximation of 2 E3:
![transrt](https://user-images.githubusercontent.com/15956904/159375840-8d75cafd-016d-4454-901b-17d377ae3a3d.png)

I think this is a first step to generalize RT to a scattering one.

HajimeKawahara:
#242

HajimeKawahara:
Also, we would include scattering.
We can use `jax.lax.linalg.tridiagonal_solve` to solve the two-stream equations.
See [an example](https://gist.github.com/HajimeKawahara/2f04acf15985f327e56fcd19eb985846).

HajimeKawahara:
in `twostream` branch, I am now developing the two stream solver (Toom/SH type).
This is the first example of the comparison with the pure absorption (using LART). It looks fine.

![Figure_1](https://github.com/HajimeKawahara/exojax/assets/15956904/c49360cf-99a1-4bfd-aa40-205b73bb79cc)

PR #410, opened 2023-09-03
(very beta) the emission w/ scattering based on fluxed-based two-stream LART solver

HajimeKawahara:
I implemented the emission w/ scattering based on the (flux-based) two-stream LART solver.

One can check that this solver agrees well with the pure absorption two-stream LART (used in HELIOS-R1 and ExoJAX1), by running `tests/integrations/unittests_log/twostream/twostream_spectrum_test.py`
You can freely change `w0` and `g` in this sample too.

## Some notes
- In this version, I assume no incident light and did not seriously the bottom boundary. So, it cannot be used for the reflection lights (i.e. only for emission w/ scattering for gaseous planets). I will think about these boundaries more seriously.
- We need some verification with other algorithms e.g. `cdisort`, etc. I include it in the future branch. 
- we plan to implement other RT solvers, such as `adding method`, 4-stream LART etc. This branch includes the basic structure for these future RT solvers (i.e. naming and `ArtEmisScat` and so on). 
- no real scattering effects such as the Rayleigh scattering, clouds have not been implemented though... 
- 
#241


HajimeKawahara:
GitHub action fails at
```
error: numpy 1.22.3 is installed but numpy>=1.22.4 is required by {'pandas'}
Error: Process completed with exit code 1.
```

HajimeKawahara:
it looks radis develop branch removes the numpy upper bound.

ykawashima (approved):
Thank you very much! I have confirmed that the code works without any problem, and the pure absorption limit agrees well with the pure absorption calculation. I also ran the code with random sets of single scattering albedo and asymmetric parameter values, confirming that the trend is reasonable.

HajimeKawahara:
Thanks a lot! because the fail of CI is due to the conflict of numpy version and pandas, and it looks radis will solve this problem in the next release, I merge this PR.

Issue #411, opened 2023-09-04
intensity-based emission spectrum (pure absorption)


HajimeKawahara:
So far, ExoJAX uses the flux-based two-stream for emission (pure absorption), also used in HELIOS-R1. 
To compare with other retrieval tools, NEMESIS, pRT, HELIOS-R2 etc, I would like to implement the intensity-based emission RT.

HajimeKawahara:
#412 for the isothermal layer

PR #412, opened 2023-09-05
intensity-based pure absorption RT solver (zero-th order)

HajimeKawahara:
I implemented the intensity-based n-stream pure absorption RT solver. This type of the RT solver have been used in NEMESIS and its successors (incl. pRT). 
#411 

You can compare the results with the current RT solver (fluxed-based two stream) by running
`tests/integration/unittests_long/pureabs/pureabs_ibased_test.py`

Figure_1

If you wanna switch your code to the ibased one, replace `rttransfer.rtrun_emis_pureabs_fbased2st(dtau, source_matrix)` (if you used the deprecated old name, this should be just `rttransfer.rtrun`) to `rttransfer.rtrun_emis_pureabs_ibased(dtau, source_matrix, mus, weights)`. `mus` and `weights` are cos theta and its weights for [Gaussian quadrature](https://en.wikipedia.org/wiki/Gaussian_quadrature), which can be computed using `rttransfer.initialize_gaussian_quadrature(nstream)`.


ykawashima (approved):
Thank you very much for the implementation! I have confirmed that the code works with the sample script. I also tested with different even nstream values.

Issue #414, opened 2023-09-16
critical bug in pressure_layer


HajimeKawahara:
#353 
Maybe I found a critical bug for the definition of the dP, in `atm.atmprof.pressure_layer_logspace`.
In v1.3-1.4.1, $`dP_n`$ was defined by 

$dP_n \equiv (1 - 10^{- dlogP/2})P_n$  --- (1), 

where $`dlogP = const`$. but it should be 

$dP_n = (10^{dlogP/2} - 10^{-dlogP/2}) P_n = k^{-1/2} - k^{1/2}$  --- (2), 

where $`k \equiv 10^{-dlogP}`$ is the pressure decrease rate, i.e. $`P_{n-1} = k P_n`$.

For an arbitrary `r = reference_point` (default; `r=0.5)`, it should be 

$dP_n = (10^{(1-r)  dlogP} - 10^{-r  dlogP}) P_n$ --- (3), 

if my brain was not completely broken by COVID-19...

Really sorry for overlooking this critical bug.

HajimeKawahara:
#414 improves the degree of the agreement with the Kawashima-san's transmission. See [here](https://github.com/HajimeKawahara/exojax/pull/359#issuecomment-1722381848).

HajimeKawahara:
#415 (master, v1.4.2) #416 (to develop)

PR #415, opened 2023-09-16
pressure_layer definition bug fix (hot fix v1.4.2)

HajimeKawahara:
Addressed #414.
This PR corrected the bug in  `atm.atmprof.pressure_layer_logspace`, and changes results for emission spectrum. So, this PR should be hotfix (v1.4.2). I will merge this change to develop later.


ykawashima (approved):
Thank you very much! I have confirmed the bug that dParr is underestimated to be kind of half and that it is now solved.

Issue #417, opened 2023-09-18
OLA for response


HajimeKawahara:
@ykawashima found in some cases, FFT in response.py limits the device memory use. I will implement the OLA (overlap-and-add) for the  response convolution.

PR #418, opened 2023-09-18
Olaresponse

HajimeKawahara:
addressed #417. 
I implemented the OLA version of `convolve_rigid_rotation` (`convolve_rigid_rotation_ola`) and `ipgauss_sampling` (`ipgauss_ola_sampling`).

The input spectrum should be reshaped to (ndiv, div_length). for instance, when the spectrum has the shape of 40000, maybe (4,10000) or (8, 5000) etc.

I hope this solves #417. I will postpone UI for them to the next PR.

### DIFF
--- a/documents/userguide.rst
+++ b/documents/userguide.rst
@@ -37,7 +37,11 @@ Radiative Transfer
 Spectral Operators
 =======================
 
-   N/A
+.. toctree::
+   :maxdepth: 1
+
+   userguide/sop.rst
+
    
 Others
 =======================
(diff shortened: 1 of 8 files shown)


HajimeKawahara:
Sample (unit tests)
- `test_convolve_rigid_rotation_ola` in `unittests/spec/response/spin_rotation_test.py`
- `test_ipgauss_ola_sampling` in `unittests/spec/response/response_test.py`

ykawashima (approved):
Thank you very much for the implementation, and I am so sorry for my very late response. I have confirmed that both sample codes work, and the OLA version produces very similar spectra calculated with convolve_rigid_rotation and ipgauss_sampling. I will try HMC calculation with the memory limit issue using this OLA feature. Thank you!

PR #419, opened 2023-09-28
intensity based linear source function (Olson and Kunasz)

HajimeKawahara:
I made the intensity based RT with the linear source approximation.
The sample is `tests/integration/unittests_long/pureabs/pureabs_ibased_linsap_test.py`.

## diff with the ibased isothermal layer
Figure_1

## diff with the flux based 2 stream
Figure_2


ykawashima (approved):
Thank you very much! I have confirmed that the sample code works as expected.

PR #422, opened 2023-09-30
r was inverse

HajimeKawahara:
I found a bug in transmission spectrum.
`radius_lower` in the integration in `rtrun_trans_pureabs` was inverse, which causes a small difference.

bug
``` python
    deltaRp2 = 2.0 * jnp.trapz(
        (1.0 - jnp.exp(-dtau_chord)) * radius_lower[::-1, None],
        x=radius_lower[::-1],
       axis=0)
```

fixed
``` python
    deltaRp2 = -2.0 * jnp.trapz(
        (1.0 - jnp.exp(-dtau_chord)) * radius_lower[:, None],
        x=radius_lower,
        axis=0)
```
The minus is because `radius_lower` is in a descending order.

### DIFF
--- a/src/exojax/atm/atmprof.py
+++ b/src/exojax/atm/atmprof.py
@@ -47,7 +47,9 @@ def pressure_layer_logspace(log_pressure_top=-8.,
     return pressure, dParr, k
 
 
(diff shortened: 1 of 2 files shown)


HajimeKawahara:
The unit tests passed (in manual...)

Issue #421, opened 2023-09-28
Simpson's rule for transmission


HajimeKawahara:
@ykawashima pointed out that Simpson's rule potentially improves the accuracy of the transmission spectra. I will think about it. 

(Currently, ExoJAX uses the trapezoidal rule)

HajimeKawahara:
- I find that we need to rewrite the chord geometric matrix by `normalized_radius_boundary = (normalized_radius_upper, 1.0)` if we do not ignore the top layer height (i=0), though the contribution is negligible.

PR #424, opened 2023-10-09
Simpson rule of the transmission chord integration

HajimeKawahara:
I implemented the Simpson integration version of the transmission spectroscopy.  #421

The demo: `tests/integration/comparison/transmission/comparison_with_kawashima_transmission.py`

I haven't checked what extent the Simpson integration version can reduce the computation time in terms of the number of the layers when fixing the uncertainty @sh-tada 
## comparison
comparison

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,6 +2,6 @@ numpyro
 jaxopt
 jax
 hitran-api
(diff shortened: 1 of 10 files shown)


ykawashima (approved):
I apologize for my late review. I have confirmed that the sample code works properly.

PR #437, opened 2023-11-24
Added Tref as an input patameter to line_strength function

ykawashima:
I have added Tref as an input parameter to [line_strength function](https://github.com/HajimeKawahara/exojax/blob/ae3f78da85ecabc5e3fb60aad0d6dacdd8d7c831/src/exojax/spec/hitran.py#L14-L37) to avoid any confusion. (I noticed that if one calls the line_strength function after OpaPremodit changed the reference temperature, one gets incorrect line strength since the change in Tref is currently not considered.)

The test script you can check with is as follows. To run this script in the develop branch to check the difference, please comment out lines 32& 48 and uncomment 33&49. Thank you!
```from exojax.utils.grids import wavenumber_grid
from exojax.spec import api
from exojax.spec import molinfo
from exojax.spec.lpf import auto_xsection
from exojax.spec.hitran import line_strength, doppler_sigma, gamma_hitran, gamma_natural, line_strength_numpy
from exojax.spec.exomol import gamma_exomol
from exojax.spec.opacalc import OpaPremodit, OpaDirect
from jax.config import config
config.update("jax_enable_x64", True)

import numpy as np
import matplotlib.pyplot as plt

nus, wav, res = wavenumber_grid(22980.0,
                                23030.0,
                                100000,
                                unit='AA',
                                xsmode="premodit")

mdb = api.MdbHitemp("/home/kawashima/database/CO/05_HITEMP2019",nus,crit=1.e-30,Ttyp=1000.,gpu_transfer=True,isotope=1)

P = 1.e-3
T = 1000.
vmr = 1.
Ppart = P * vmr
Mmol = molinfo.molmass("CO")

logsij0 = np.log(mdb.line_strength_ref)
sigmaD = doppler_sigma(mdb.nu_lines,T,Mmol)
qt = mdb.qr_interp(mdb.isotope, T)
gammaL = gamma_hitran(P,T, Ppart, mdb.n_air, mdb.gamma_air, mdb.gamma_self) + gamma_natural(mdb.A)
Sij = line_strength(T,logsij0,mdb.nu_lines,mdb.elower,qt, mdb.Tref)
#Sij = line_strength(T,logsij0,mdb.nu_lines,mdb.elower,qt)                                                                                                          
xsv0 = auto_xsection(np.array(nus),mdb.nu_lines,sigmaD,gammaL,Sij,memory_size=30)

opa = OpaPremodit(mdb=mdb,
                  nu_grid=np.array(nus),
                  diffmode=2,
[...]

PR #440, opened 2023-11-24
vmap index for  line strength

HajimeKawahara:
Some correction for vmap(line_strength).
I think we should put `None` instead of 0 as the Tref argument in vmapped line_strength
Yui's code in https://github.com/HajimeKawahara/exojax/pull/437 is now included in tests/integration/api/api_premodit_to_direct.py
Looks fine but confirm it?


ykawashima (approved):
Thank you very much for changing Tref argument from 0 to None (sorry, I did not notice). I have also confirmed that my code added works. Thank you!

PR #445, opened 2023-12-02
Refactoring testdata

HajimeKawahara:
removed pickles and its relationals because we do not need them anymore except for VALD.


HajimeKawahara:
I love refactoring

Issue #453, opened 2024-01-10
optelower does not work for HITEMP


HajimeKawahara:
`optgrid.optelower` was designed only for ExoMol, and actually does not work for HITEMP.

PR #454, opened 2024-01-10
optelower for HITEMP

HajimeKawahara:
addressed #453 

See `test/integration/unittest_long/premodit/optgrid_test.py`


HajimeKawahara:
this is a small change so self-merge

PR #457, opened 2024-01-11
Scripts for the comparison with petitRADTRANS

ykawashima:
This is just to add the script for the comparison with petitRADTRANS.


HajimeKawahara (approved):
Thank you so much!

Issue #458, opened 2024-01-12
Error in MdbExomol when code level in .broad file is 'a1'


sh-tada:
I encountered an error while using `MdbExomol`. This error seems to occur when the code level in the .broad file "a1".
It appears that this issue is related to a bug in the [exomolapi.py](https://github.com/radis/radis/blob/develop/radis/api/exomolapi.py) file in radis. I am planning to submit a Pull Request to radis to address this issue.

- Error messages
```
Traceback (most recent call last):
  File "/home/tada/anaconda3/envs/exojax/lib/python3.10/site-packages/ExoJAX-1.4.3-py3.10.egg/exojax/spec/api.py", line 190, in activate
    self.compute_broadening(self.jlower.astype(int), self.jupper.astype(int))
AttributeError: 'MdbExomol' object has no attribute 'compute_broadening'. Did you mean: 'computebroadening'?

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/tada/exojax/spectrum.py", line 78, in <module>
    mdb_H2O = MdbExomol(".db_ExoMol/H2O/1H2-16O/POKAZATEL/", nu_grid, gpu_transfer=False)
  File "/home/tada/anaconda3/envs/exojax/lib/python3.10/site-packages/ExoJAX-1.4.3-py3.10.egg/exojax/spec/api.py", line 141, in init
    self.activate(df)
  File "/home/tada/anaconda3/envs/exojax/lib/python3.10/site-packages/ExoJAX-1.4.3-py3.10.egg/exojax/spec/api.py", line 192, in activate
    self.set_broadening_coef(df[mask], add_columns=False)
  File "/home/tada/exojax/exojax/src/radis/radis/api/exomolapi.py", line 1285, in set_broadening_coef
    jlower_max=np.max(df["jlower"]),
  File "/home/tada/anaconda3/envs/exojax/lib/python3.10/site-packages/numpy/core/fromnumeric.py", line 2810, in max
    return _wrapreduction(a, np.maximum, 'max', axis, None, out,
  File "/home/tada/anaconda3/envs/exojax/lib/python3.10/site-packages/numpy/core/fromnumeric.py", line 86, in _wrapreduction
    return reduction(axis=axis, out=out, **passkwargs)
TypeError: Expression.max() got an unexpected keyword argument 'axis' 
```

HajimeKawahara:
merged into develop in radis

Issue #408, opened 2023-08-04
Guarantee compatibility with Radis, the common API


HajimeKawahara:
See https://github.com/radis/radis/issues/597

PR #462, opened 2024-01-13
uncomment install_radis_develop() in setup.py?

ykawashima:
I just thought it might be good to uncomment ``install_radis_develop()`` in setup.py for the time being as the current develop branch keeps installing radis==0.15, causing the issue reported in #458 (radis==0.14 does not cause this error).

But this is just my opinion, and this pull request is just in case you found it useful, too. Please don't hesitate to ignore it otherwise.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -86,7 +86,7 @@ def reinstall(package):
     subprocess.check_call([sys.executable, "-m", "pip", "uninstall", "-y", package])
     subprocess.check_call([sys.executable, "-m", "pip", "install", package])
 
-install_radis_develop()
+#install_radis_develop()
 uninstall('vaex-core')
 uninstall('vaex-astro')
 uninstall('vaex-jupyter')


HajimeKawahara:
This is a troubling issue. Currently, several issues have been resolved in the develop version of `radis`, but if we apply the `radis` master branch (currently `radis==0.14`) to `exojax` develop branch, these issues will remain unresolved in the `exojax` develop branch. 

In the short term, since @sh-tada is planning to submit a PR to `radis` to resolve issue #458, this problem will be resolved temporarily once it gets merged into `radis` develop branch, but...

What do you think about this issue? @erwanp @minouHub

HajimeKawahara:
See also #408

minouHub:
I understand that this issue is also discussed in https://github.com/radis/radis/issues/597 and that some tests would help with this issue. However, it would be more convenient that the develop version of any code (including `exojax`) relies on the master version of `radis`. For instance, I would not work on a new PR that depends on a temporary version of `Numpy`, as it is likely to change in the following weeks.
In essence, I agree with this PR.

HajimeKawahara:
Thanks, understood. In that case, I also agree with the direction of this PR. I would like to propose the following, what do you think? @minouHub @ykawashima @erwanp 

1. Since the current `exojax/develop` depends on `radis/develop`, after the release of `radis==0.15`, we will change `exojax/develop setup.py` to install `radis==0.15`. 
2. We will add the unittests being discussed in [radis #597](https://github.com/radis/radis/issues/597) to `radis/develop`. 

By the way, is there a plan to merge `radis/develop` into `radis/master` and release `radis==0.15` anytime soon?

ykawashima:
The plan sounds good to me! Thank you.

minouHub:
@erwanp, we need to talk about merging dev into master :)

HajimeKawahara:
postponed to v1.6

HajimeKawahara:
I am working on this issue in #500 so I close this PR.

PR #464, opened 2024-01-15
Fix opachord.py to avoid nan in the gradient of ArtTransPure model 

sh-tada:
This pull request resolves the issue (#463) of NaN values in the gradient of the `ArtTransPure` transmission spectra model with respect to temperature, gravity_btm, radius_btm, and mean molecular weight. The NaN values originated from the `jnp.sqrt` function in [`exojax/spec/opachord.py`](https://github.com/HajimeKawahara/exojax/blob/develop/src/exojax/spec/opachord.py).

To address this, I implemented `jnp.tril` before the `jnp.sqrt` operation. This modification zeros out irrelevant values, ensuring proper gradient calculations. Setting k=-1 in `jnp.tril` also zeroes the diagonal.

Fixes #463

PR #465, opened 2024-01-15
Refactoring transmission related codes

HajimeKawahara:
Just refactoring after #464


sh-tada:
Other than the comment above, it seems there are no issues.
The revised program worked without any issues.

PR #472, opened 2024-02-02
added nstream=2 just in case in the test scripts for the comparison with petitRADTRANS

ykawashima:
I have just added ``art.nstream = 2`` in the test scripts for the comparison with petitRADTRANS when changing from ``art.rtsolver = ibased`` to ``art.rtsolver = fbased2st``. This is just to avoid possible confusion, and I have confirmed that there would not be any change for the results since the nstream parameter is not used for ``art.rtsolver = fbased2st``. Thank you.


HajimeKawahara (approved):
Thanks!

PR #482, opened 2024-04-02
Change the order of calculations for 32-bit mode

### LINKED ISSUE #473: The mmw derivative of the transmission spectra model becomes NaN when using 32-bit
I encountered an issue where the derivative of the transmission spectra model (`ArtTransPure`) with respect to the mean molecular weight (mmw) results in NaN (Not a Number) when operating in a 32-bit environment. This problem does not occur in a 64-bit setting.

You can reproduce this issue by running [exojax/tests/integration/unittests_long/transmission
/transmission_grad_test.py](https://github.com/HajimeKawahara/exojax/blob/develop/tests/integration/unittests_long/transmission/transmission_grad_test.py) after commenting out [`config.update("jax_enable_x64", True)`](https://github.com/HajimeKawahara/exojax/blob/dc86768c0f03b66b4fda0d95cb1680c2b51d6dcb/tests/integration/unittests_long/transmission/transmission_grad_test.py#L20C1-L20C38) and adding `allow_32bit=True` to [OpaPremodit](https://github.com/HajimeKawahara/exojax/blob/dc86768c0f03b66b4fda0d95cb1680c2b51d6dcb/tests/integration/unittests_long/transmission/transmission_grad_test.py#L45).

sh-tada:
I found that we can prevent NaNs in gradients that occur only in 32-bit mode by modifying the order of calculations. 
This may be because it is better to perform computation with 64-bit variables before applying 32-bit variables to minimize rounding errors.

Fixes #473

### DIFF
--- a/src/exojax/atm/atmprof.py
+++ b/src/exojax/atm/atmprof.py
@@ -153,7 +153,9 @@ def gh_product(temperature, mean_molecular_weight):
     Returns:
         gravity x pressure scale height cm2/s2
     """
-    return kB * temperature / (m_u * mean_molecular_weight)
+    return (
+        kB * temperature / m_u / mean_molecular_weight
+    )  # Apply mmw (jnp array) last to minimize rounding errors in 32bit mode.
 
 
 def pressure_scale_height(gravity, T, mean_molecular_weight):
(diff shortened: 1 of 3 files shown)


HajimeKawahara (approved):
Thanks!

Issue #488, opened 2024-06-03
Multiplying `10**gamSta` by electron number density


code29563:
I see that the Stark broadening parameter `10**gamSta` in the `gamma_` functions in atomll.py isn''t multiplied by the electron number density although [the Vald3 format page](https://www.astro.uu.se/valdwiki/Vald3Format) specifies their units as '(s*Ne)^-1' 'where Ne is the electron number density'. Is this a mistake?

HajimeKawahara:
Thanks! @chonma0ctopus Can you check it?

chonma0ctopus:
Thank you for pointing it out! I will check it in a day.

chonma0ctopus:
Sorry for the confusion. It was not the version I processed this time, it was giving the error before I edited it. I haven't been able to pinpoint at which point in development the cause was.

I thought it might be at Issue #488, but it wasn't; at 49b9e3283ea64b3cee8197a3ebe058b442b9eaad we already get an AssertionError.

Do you happen to know any smart way to track down at what point the difference occurred? Or, do you think there is no need to track it and rather I can just update the assert statement?

PR #489, opened 2024-06-07
Bugfix: gamma for Stark broadening of metal lines

chonma0ctopus:
As pointed out in #488, the Stark broadening parameter is in a unit of '(s*Ne)^-1' but was mistakenly not multiplied by Ne (electron number density). Also, a temperature-dependent factor (`*(T/10000.)**(1/6)`; c.f. https://ui.adsabs.harvard.edu/abs/2006AIPC..874...35G/abstract) was forgotten. I fixed them.


HajimeKawahara (approved):
Thanks!

Issue #487, opened 2024-05-26
`qt_K` -> `qr` in Forward_modeling_for_Fe_I_lines_of_Kurucz.ipynb


code29563:
I think something missing from [this notebook](https://github.com/HajimeKawahara/exojax/blob/master/documents/tutorials/Forward_modeling_for_Fe_I_lines_of_Kurucz.ipynb) is to divide `qt_K` determined in cell 5 by `adbK.QTref_284[mask]` before passing to `SijT` (or now `line_strength`) in the next cell, as the function requires `qr: partition function ratio qr(T) = Q(T)/Q(Tref)` for that argument. Could you confirm?

HajimeKawahara:
Thanks. @chonma0ctopus Can you also check this?

HajimeKawahara:
Working on this issue at `kurucz_bug` branch

chonma0ctopus:
I apologize for the bug I left behind that has caused you all trouble. After a quick look, I think @code29563 has made the right point. I will take the time to look at it more closely to modify them next Sunday night.

HajimeKawahara:
@chonma0ctopus I reached Cell 5 by modifying Cell 1-4.  Can you take over `kurucz_bug` branch?

chonma0ctopus:
Understood.

HajimeKawahara:
@chonma0ctopus Have you identified the cause of the error?

chonma0ctopus:
Sorry for the further delay. I've finally taken the time to manage the issues in the notebook: https://github.com/chonma0ctopus/exojax/tree/kurucz_bug

I see similar problems in a few other files (e.g., `Tref` assignment issues on `line_strength`, `qt` used instead of `qr`, and some deprecated functions...). I'm planning to tackle these too:
- lpf.py
- opacity_Fe_test.py
- Reverse_modeling_with_VALD_using_MODIT.ipynb
- Forward_modeling_for_Fe_I_lines_of_Kurucz.rst
- Forward_modeling_for_metal_line.rst
(There might be more, I'll keep an eye out!)

Quick question: Should I go ahead and merge the notebook fixes into your develop branch first? Or wait until I've sorted out the other files too?

HajimeKawahara:
@chonma0ctopus Thanks! Choose one that suits you best!

chonma0ctopus:
Got it! I'll work on the other files first next weekend.

chonma0ctopus:
[...]

PR #499, opened 2024-07-24
Correct Q(T) to Q(T)/Q(Tref)

chonma0ctopus:
Correct the wrong implementation of Q(T) ratio pointed out in #487


HajimeKawahara (approved):
Worked at least my environment.

PR #500, opened 2024-08-03
Matches the installation requirements for radis 0.15

HajimeKawahara:
Since `radis 0.15` has been released, review the requirements.txt and setup.py of ExoJAX accordingly.
I aim to support Python >=3.11 too #502


HajimeKawahara:
#462

minouHub:
Btw. Radis=0.15 is compatible up to python=3.11. The only caveat is that for python>=3.10, vaex is not installed and we only use pandas. Waiting for vaex to upgrade 🙂

Issue #502, opened 2024-08-03
support for python 3.11 


HajimeKawahara:
basically same as https://github.com/radis/radis/issues/653 and https://github.com/radis/radis/issues/656

We will make vaex an option.

HajimeKawahara:
close with #500

erwanp:
For information Vaex support for Python 3.11; 3.12 was merged https://github.com/vaexio/vaex/pull/2331

PR #506, opened 2024-08-07
api.py for pytables

HajimeKawahara:
if choosing `pytables` as radis engine, columns was empty because `vaex` is lazy I/O. So I fixed it and now the user can manually set the radis engine in Mdb(Exomol, Hitemp, Hitran).


HajimeKawahara:
Check the integration test: exojax/tests/integration/api

HajimeKawahara:
See https://secondearths.sakura.ne.jp/exojax/develop/userguide/api.html for the doc

HajimeKawahara:
#502

PR #530, opened 2024-09-08
radis consistency

HajimeKawahara:
added the same unit test for the common API to check if the proper version of RADIS is installed.


HajimeKawahara:
after merging https://github.com/radis/radis/pull/693 to radis/develop, the unit test will pass.

erwanp:
https://github.com/radis/radis/pull/693 was merged

PR #538, opened 2024-09-19
updated the petitRADTRANS comparison codes due to the update of petit…

ykawashima:
This is just to reflect the update of petitRADTRANS for the petitRADTRANS comparison codes. Also, I changed nstream since petitRADTRANS seems to use nstream=6 (cf. Eq. (3) of Molliere et al. 2019 and 
https://gitlab.com/mauricemolli/petitRADTRANS/-/blame/master/petitRADTRANS/radtrans.py#L211). Thank you.


HajimeKawahara (approved):
Thank you so much!

Issue #559, opened 2025-01-13
[opart] layer-wise computation for CIA/Rayleigh scattering/H-


HajimeKawahara:
related to #546

HajimeKawahara:
### CIA
- `spec.layeropacity.singlelayer_optical_depth_CIA`

#### demo
- `get_started_opart`

### H- 
- `spec.layeropacity.singlelayer_optical_depth_hminus`
- `spec.hminus.log_hminus_continuum_single`

### Rayleigh scattering

PR #565, opened 2025-01-26
Opart continuum

HajimeKawahara:
addresses #559 

See [here](https://secondearths.sakura.ne.jp/exojax/develop/tutorials/get_started_opart.html)


HajimeKawahara:
self merge

PR #586, opened 2025-03-01
n_Texp grid margin bug

HajimeKawahara:
Bugfix for the following bug:
The `nextafter` function may not work for the endpoints of `spec.set_ditgrid.ditgrid_linear_interval`.

This bug may cause an inconsistency between the shape of `multi_index_uniqgrid` and the length of `n_Texp_grid`. Specifically, when some lines have values equal to the endpoints of `n_Texp_grid`, an extra dimension is unintentionally added to `multi_index_uniqgrid`. Consequently, in the following part of the `spec.premodit.unbiased_ngamma_grid` code, a nonexistent index may be referenced.

```python
n_Texp_g = n_Texp_grid[multi_index_uniqgrid[:, 1]]
```
That is, even when `len(n_Texp_grid) = 2`, the code may attempt to access `n_Texp_grid[2]`. Normally, this would result in an `IndexError`, like
```sh
  File "/home/kawahara/exojax/src/exojax/spec/premodit.py", line 126, in xsvector_nu_open_zeroth
    ngamma_grid = unbiased_ngamma_grid(
  File "/home/kawahara/exojax/src/exojax/spec/premodit.py", line 1495, in unbiased_ngamma_grid
    n_Texp_g = n_Texp_grid[multi_index_uniqgrid[:, 1]]
IndexError: index 2 is out of bounds for axis 0 with size 2
```
but surprisingly, when the function is JIT-compiled, no error occurs. Instead, it stores the value of `n_Texp_grid[-1]`. As a result, no numerical error is raised, but the memory usage for `lbd_coeff` increases. When `len(n_Texp_grid) = 2`, the memory consumption increases by a factor of 1.5.  

This PR fixes the root cause in `set_ditgrid` and adds a function (`spec.premodit.check_multi_index_uniqgrid_shape`) to check the consistency among `multi_index_uniqgrid`, `n_Texp_grid`, and `ngamma_ref_grid`.  

Further investigation will be conducted on why JAX does not raise an `IndexError`, and an issue will be submitted to the JAX repository.

### DIFF
--- a/src/exojax/spec/premodit.py
+++ b/src/exojax/spec/premodit.py
@@ -23,7 +23,6 @@ from exojax.spec.lbd import lbd_coefficients
 from functools import partial
 
 
-
 @partial(jit, static_argnums=14)
 def xsvector_open_zeroth(
     T,
@@ -184,6 +183,7 @@ def xsvector_open_second(
     )
     return xs
 
+
 @partial(jit, static_argnums=14)
 def xsmatrix_open_zeroth(
     Tarr,
@@ -238,6 +238,7 @@ def xsmatrix_open_zeroth(
         Slsd, R, nsigmaD, nu_grid_extended, log_ngammaL_grid, filter_length_oneside
     )
     return xsm
+
 
 @partial(jit, static_argnums=14)
 def xsmatrix_open_first(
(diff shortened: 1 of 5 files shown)


HajimeKawahara:
See also #590 (changed dtype for `np.nextafter`)

PR #596, opened 2025-04-28
add document on using cloudpickle

sh-tada:
* Adds `opa_pickle.rst`
  - Shows a minimal example using `cloudpickle.dump / load`
* No library code changes


HajimeKawahara:
It appears that these unittest errors are related to [the recent release](https://github.com/radis/radis/pull/819) of RADIS....

PR #598, opened 2025-05-12
Set custom JVP for sqrt in opachord to enable forward mode differentiation of transmission

sh-tada:
Fixed NaN issue in forward mode differentiation of the transmission spectrum by adding a custom JVP for sqrt in opachord.py.

Changes:
Implemented custom JVP for sqrt to handle division by zero.

Note:
Without [at]jit for chord_geometric_matrix_lower and chord_geometric_matrix, both branches of where may be evaluated, causing division by zero when x = 0.

### DIFF
--- a/src/exojax/spec/opachord.py
+++ b/src/exojax/spec/opachord.py
@@ -4,7 +4,22 @@ from jax import jit, custom_jvp
 
 @custom_jvp
 def safe_sqrt(x):
-    # Avoids division by zero in the derivative of sqrt
+    """Element-wise square root with a *safe* derivative.
+
+    The custom JVP sets the gradient to zero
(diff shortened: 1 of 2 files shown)


src/exojax/spec/opachord.py
```diff
@@ -1,5 +1,27 @@
...
+def _(primals, tangents):
+    (x,), (t,) = primals, tangents
+    y = safe_sqrt(x)
+
+    # Derivative: t / (2 * y), with zero for x <= 0 to prevent NaN
+    dx = jnp.where(x > 0, t / (2 * y), 0.0)
+    return y, dx
+
+
+# Note:
```
HajimeKawahara, line 21: Can you write these comments in the function (for the API)?

HajimeKawahara:
Thanks!

Issue #619, opened 2025-07-31
nan values in cross-section calculation with 32-bit mode


ykawashima:
I noticed that when using 32-bit precision, the cross-section calculation returns nan values.  It appears that [fftvalvk](https://github.com/HajimeKawahara/exojax/blob/3014a8b2accccbf181fd19aba6c6c46bd8359010/src/exojax/opacity/_common/profconv.py#L114) results in nan values in 32-bit mode. I have confirmed this behavior with both ``ExoJAX==2.1`` and ``ExoJAX==2.0``, as well as ``JAX==0.4.30`` and ``JAX==0.6.2``. Do you have any idea what might be causing this? Thank you.

P.S. I realized that adding crit=1.e-30 avoids the issue. Small values might be causing the problem.

A sample script is shown below:
```
from jax import config
# config.update("jax_enable_x64", True)                                                                                        

from exojax.utils.grids import wavenumber_grid
nu_grid, wav, resolution = wavenumber_grid(22920.0, 23000.0, 3500, unit="AA", xsmode="premodit")

from exojax.database.api  import MdbExomol
mdb = MdbExomol("/home/kawashima/media/ssds/database/CO/12C-16O/Li2015", nurange=nu_grid)

from exojax.opacity import OpaPremodit
opa = OpaPremodit(mdb, nu_grid, auto_trange=[500.0, 1500.0], dit_grid_resolution=1.0, allow_32bit=True)

P = 1.0  # bar                                                                                                                
T_1 = 500.0  # K                                                                                                              
xsv_1 = opa.xsvector(T_1, P)  # cm2                                                                                           
print(xsv_1)
```

HajimeKawahara:
Thank you. I will check it. @ykawashima

PR #620, opened 2025-08-04
bug fix of the overflow of the line strength when 32 bit mode

HajimeKawahara:
This PR addresses #619

The reason of `nan` in cross section using the 32-bit mode was
- `elower` was converted to `jnp.array` (float32, when 32-bit mode) when `gpu_transfer=True` (default) in `mdb`. 
- Also, `qr_interp` was float32.
This makes the overflow of `line_strength` and `Slsd`.

## Solutions
- removes `jnp.array`-nization in `generate_jnp.arrays` in `mdb`'s. 
- makes the numpy version of  `QT_interp` and `qr_interp` 

## refactoring in near future
we need further refactoring of `databases/api.py`

### DIFF
--- a/src/exojax/database/api.py
+++ b/src/exojax/database/api.py
@@ -56,6 +56,9 @@ def _set_engine(engine):
     print("radis engine = ", engine_selected)
     return engine_selected
 
+def _checkf64(instance, name):
+    if instance.dtype != np.float64:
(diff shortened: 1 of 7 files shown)


ykawashima (approved):
Thank you very much! I have confirmed that the issue has been resolved.

PR #630, opened 2025-09-04
Detach OpaPremodit from mdb to reduce memory footprint

sh-tada:
This PR refactors `OpaPremodit` so it no longer keeps a long-lived reference to `mdb`. During initialization, `OpaPremodit` extracts only the data it needs from `mdb` and then releases the handle.

### Note
- Accessing `opa.mdb.<attr>` is no longer supported.
- These changes may increase initialization time. In particular, we observe higher runtime in `opacity._common.lsd.npadd3D_multi_index` within `opacity.premodit.premodit.generate_lbd` within `opacity.initspec.init_premodit` on this branch.


HajimeKawahara (approved):
Thanks!

Issue #612, opened 2025-06-24
refactoring tests


HajimeKawahara:
- Some files in `testdata` directory are not used in unittests

PR #679, opened 2026-01-25
removed unused testdata

HajimeKawahara:
addresses #612


HajimeKawahara:
`database.hargreaves.api._convert_to_exomol`

Fixes a failure under pandas ≥3.0 (Copy-on-Write), where modifying a NumPy array obtained via .values can raise
ValueError: assignment destination is read-only.

Missing values are now filled at the pandas level before conversion to NumPy, avoiding in-place modification of a shared buffer.

PR #681, opened 2026-03-04
Added Me, gE, and gravity_earth

ykawashima:
This pull request just adds Earth parameters (values from astropy) and the gravity_earth function. Thank you.

Test code:
```
from exojax.utils.astrofunc import gravity_earth
import  exojax.utils.constants as const

gravity = gravity_earth(1.0, 1.0)
print(gravity)

ME = const.ME
RE = const.RE
gE = const.gE
print(ME, RE, gE)
```


HajimeKawahara (approved):

HajimeKawahara:
Thank you! merged.

PR #726, opened 2026-07-23
Warn for insufficient instrumental profile vrmax (check_vrmax)

### LINKED ISSUE #725: warning for insufficient vr_array
For low spectral resolution, the default vr_array length is not enough. We would warn for this case.

HajimeKawahara:
## Summary

- add `SopInstProfile.check_vrmax` to validate the velocity range used
by a Gaus
sian instrumental profile
- warn automatically from `SopInstProfile.ipgauss` when `vrmax` covers
fewer tha
n five Gaussian standard deviations
- keep the check independent of the bin spacing of `nu_grid`
- document how to choose a fixed `vrmax` before JIT-compiled NumPyro
NUTS calcul
ations

## Motivation

For low instrumental resolving power, the Gaussian standard deviation
can become
  comparable to or larger than the configured `vrmax`. The kernel is
then silentl
y truncated by the fixed velocity range and renormalized, producing a
narrower i
nstrumental profile than requested.

The new check uses

\[
\mathrm{vrmax} \geq 5\beta_\mathrm{inst},
\qquad
\beta_\mathrm{inst}
= \frac{c}{2\sqrt{2\ln 2}\,R_\mathrm{inst}}.
\]

The model-grid resolution is intentionally not used in this truncation
check. Sa
mpling adequacy is a separate concern.

## Usage

```python
from exojax.postproc.specop import SopInstProfile
from exojax.utils.instfunc import resolution_to_gaussian_std

Rinst = 3000.0
beta_inst = resolution_to_gaussian_std(Rinst)

# Fix the velocity-grid shape before JIT/NUTS compilation.
vrmax_inst = 5.0 * beta_inst
sop_inst = SopInstProfile(nu_grid, vrmax=vrmax_inst)

# Optional explicit preflight check. ipgauss also performs this check.
sop_inst.check_vrmax(beta_inst)
spectrum_inst = sop_inst.ipgauss(spectrum, beta_inst)
```

For the usual case where `Rinst` is fixed outside a NumPyro model, this
keeps th
e velocity-grid shape constant and does not introduce recompilation
during NUTS.
  Traced standard deviations are skipped to avoid runtime host callbacks.

## Validation

- `pytest -q tests/unittests/postproc/response`: 20 passed
- `ruff check src/exojax/postproc/specop.py
tests/unittests/postproc/response/re
sponse_test.py`
- `git diff --check`

Closes #725.

### DIFF
--- a/documents/userguide/sop.rst
+++ b/documents/userguide/sop.rst
@@ -7,7 +7,7 @@ In the post-radiative transfer, the observed spectrum differs from the raw spect
 For instance, it might experience rotational broadening due to the planet's rotation, wavelength shifts 
 due to differences in line-of-sight velocities, or the influence of the instrument's profile (IP). 
 In ExoJAX, these responses to the spectrum are termed the "Spectral Operator" (``sop``). 
-Within the ``spec.specop`` module, classes like ``SopRotation`` and ``SopInstProfile`` allow for the easy handling of these responses.
+Within the ``postproc.specop`` module, classes like ``SopRotation`` and ``SopInstProfile`` allow for the easy handling of these responses.
 
 
 SopRotation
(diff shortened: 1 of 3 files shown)


ykawashima (approved):
Thank you very much for the implementation! I have confirmed that a warning is issued when the velocity grid is insufficient and that all three newly added tests pass.

HajimeKawahara:
Thank you! @ykawashima

PR #798, opened 2026-09-09
Fix RV boundary and roundoff handling in benchmark validation

HajimeKawahara:
Finite-difference checks in the DiffGrid NUTS benchmark can cross a linear RV interpolation boundary and report a misleading derivative mismatch. Very small perturbations can also disappear through floating-point rounding and falsely pass a check.

Include RV sampling cells and exact knots in the smooth-region labels, preserve one-sided diagnostics for boundary crossings, and reject unresolved perturbations in input, physical-parameter, Doppler, and logit-transformed coordinates. Extend the default step sequence through `1e-6` and `1e-7` so nearby smooth points can satisfy the existing two-adjacent-step criterion without relaxing tolerances.

Regression tests cover forward and potential coordinates, exact knots, nearby crossings, custom step sequences, and partial or transformed roundoff.

## Validation

- CPU: 71 tests passed across `test_benchmark_metrics.py`, `test_diffgrid_nuts_validation.py`, and `tests/unittests/postproc/test_response.py`.
- `git diff --check` passed.

### DIFF
--- a/tests/benchmark/benchmark_metrics.py
+++ b/tests/benchmark/benchmark_metrics.py
@@ -113,14 +113,18 @@ def directional_check(
     tolerance=1e-3,
     in_domain=None,
     region=None,
+    stencil_resolved=None,
 ):
     """Compare JAX JVP with central differences at two adjacent supplied steps.
 
     ``in_domain`` rejects endpoints outside the allowed physical/prior domain.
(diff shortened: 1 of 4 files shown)


HajimeKawahara:
See #731 for the results

Issue #807, opened 2026-09-16
Map current database loading paths and backend boundaries


HajimeKawahara:
# Current database loading paths

ExoJAX currently loads spectral data through RADIS, optional PyExoCross readers, and native ExoJAX readers. The ownership of downloading, parsing, caching, and physical normalization differs between these paths. This inventory makes those boundaries explicit after the ExoMol backend and ExoAtom additions.

Code snapshot: [174295c0](https://github.com/HajimeKawahara/exojax/tree/174295c0bcfe01e3eb564262ce00ee86e3b75366), inspected on 2026-09-16. Related: #804, #805, #806, and the provider-layer discussion in #666. Source links below are pinned to this snapshot.

## How to read the map

- **Source** identifies the dataset: ExoMol, ExoAtom, HITRAN, NIST ASD, etc.
- **Reader/backend** identifies the code that parses it: RADIS, PyExoCross, or ExoJAX. Only `MdbExomol` currently exposes the `backend=` choice.
- **Engine** selects a storage/DataFrame implementation where supported. For example, `engine="vaex"` does not select a different scientific database or replace RADIS with PyExoCross.
- **Normalization** prepares strengths, partition functions, broadening information, filtering, and NumPy/JAX arrays for the opacity layer. ExoJAX assembles source-specific opacity inputs, reusing normalization supplied by its backend where applicable.

## Molecular line data

<img width="524" height="291" alt="Image" src="https://github.com/user-attachments/assets/0b04864a-78e6-4937-b26a-74c83fee5ec0" />

| Entry point | Acquisition and reader | Local reuse / cache | Output and notable behavior |
| --- | --- | --- | --- |
| [`MdbExomol(..., backend="radis")`][exomol] **default** | RADIS ExoMol manager handles acquisition and parsing through the [RADIS adapter][radis]. | RADIS parsed-file caches and local source files; `engine=None` lets RADIS choose its engine. | RADIS metadata, reference strengths, and broadening coefficients; ExoJAX filtering, activation, partition interpolation, and line arrays. |
[...]

PR #808, opened 2026-09-16
docs: summarize molecular and atomic database loading paths

HajimeKawahara:
Refresh the existing database overview with concise molecular and atomic database tables and the two diagrams supplied in #807. Link to existing usage guides and clarify RADIS/PyExoCross selection, local versus downloaded inputs, and NIST/ExoAtom broadening requirements.

The diagrams are stored locally without modification. The page is already the first entry in the User Guide's Databases section.

Validation: isolated Sphinx HTML build of the overview with warnings treated as errors, existing documentation link targets and image assets checked, and `git diff --check` passed.

Related to #807.


HajimeKawahara:
#807

